PR #15, opened 2019-02-07
VS2019, Parallelization, SSE2 for FillAlphaMashBody on x64 and other fixes

jesec:
First, this PR is NOT designed to be merged directly. It might be better to cherry-pick from it. 

I encountered some performance issues when I play a video with fancy subs few days before so I checked the source of this project out. 

By making these changes, I got an approx. 25% solid performance improvement (measured via automatic unit_test) with the same build environment and configurations (re-targeted to VS2019 and Windows SDK 10.0). Subjectively there is ~70% improvement in my 4K, madVR, NGU very high setup. 

Please review these changes. Thank you. 

Binaries of my build are available at https://github.com/jesec/xy-VSFilter/releases.
Note that older and 32-bit systems may not be supported.


madshi:
Awesome, thank you for contributing!  :)  Hopefully cyberbeing can have a look at this when he finds some time?

jesec:
Hi @madshi . First, thanks for your madVR. :D 
I also played with libass with directshow filter today. I noticed there are some very SERIOUS performance improvement compared to xySubFilter. dsfilter with libass backend can handle very demanding sub in 4K real time while even my optimized xySubFilter needs to rely on madVR queue to avoid sudden frame drops sometimes. It is a bit shocking to me because last time I tried libass (AssFilterMod binary release), its performance is a p* of s* compared to xySubFilter. Note that this is subjective comparison. I will probably look into this further and do more automatic unit test comparisons when I am free. 

The dsfilter for libass is not as feature complete as xySubFilter though. Some features are missing (eg. open external sub). I will probably take up the development if I confirm there are real performance improvement and if I have time.

madshi:
That'd be wonderful !!

Cyberbeing:
@jesec 
One thing you need to realize about xy-VSFilter/XySubFilter, is that the majority of the speed improvements compared to VSFilter 2.39 came from the addition of various caches to hide VSFilter's poor scan convert performance. The SSE2 optimizations which were done for blur, border, alpha, script parsing, and other things only help so much alone. When caching fails, performance falls apart. 

I've long speculated is caused by performance bottlenecks in VSFilter's inefficient use the GDI32 backend to achieve sub-pixel positioning for fonts and vectors. Unlike the Libass, VSFilter does this in a rather hacky way and requests everything from GDI32 at 8x8 supersampled size, performing layout and scan convert with fixed-point logic. Moving 3D Transforms in particular can suffer heavily from this, since at worse each frame gets rendered entirely from scratch and snapping to the nearest sub-pixel(?) which can sometimes even make motion jerky.

Needless to say the current logic becomes can become extremely slow on 4K and above even on modern CPUs (i.e. a fullscreen 3840x2160 object would be 30720x17280 internally...). This is a legacy from VSFilter's early Win98 days when 240p was the norm and maybe this didn't seem as insane as it does now. It is also the source of VSFilter's memory usage issue with large objects.

At one point in time, the original dev had planned to write a new floating-point scan converter from scratch (or possibly create something similar to ftgrays from freetype) but unfortunately never got around it it before the project died. Either doing that, or replacing GDI32 entirely with something like Direct2D/DirectWrite or even Freetype is probably what would need to be done to keep performance acceptable into the future if multi-threading is not possible.

Cyberbeing:
@jesec 
As for the pull request itself, would you able to break-down the performance improvement of each commit you think it noteworthy? Personally I'm still running Win7 SP1 x64 on a CPU which only support AVX1, so in particular I'm curious how much performance (if any) you are losing if instead of Win10/AVX2 you created a Win7/SSE2 or Win7/AVX1 compatible build.

I'd also like to confirm if your changes in e505ef1 are equivalent in terms of fixing the rounding error, originally committed in a5028bc0 , 09d981b4 , 0536e37e , 97bc77dc ?

At quick look at the rest seems okay, I'll test it out later on tonight or tomorrow.

madshi:
I've recently looked into using GDI+ myself (not for subtitles, though) and it works quite well for me. But I've not benchmarked it. It does support subpixel positioning, but it's reportedly quite a bit slower than GDI32. I don't know, maybe Direct2D/DirectWrite would perform better, by using GPU hardware support or something, but I've never used that yet, so I couldn't say.

Cyberbeing:
From what I remember about GDI+, it uses resolution independent rendering, while VSFilter is written around GDI32 which is resolution dependent, potentially making it not so much a drop in replacement. Also in terms of pure throughput, I was under the impression that GDI32 should actually be faster GDI+, but that would need to be tested for our particular use-case. A quick google brought up https://medium.com/smooth-as-butter/performance-of-image-drawing-gdi-vs-gdi-8ca767f37a62 which suggests the GDI+ Draw function is extremely slow but the others are fine.

In terms of CPU-based implementations, I believe Google's open-source Skia rendering library is supposed to be faster than GDI+ and would likely be the preferred option to target if Direct2D/DirectWrite didn't seem practical.

Youka:
> In terms of CPU-based implementations, I believe Google's open-source Skia rendering library is supposed to be faster than GDI+...

Was [cairo](https://www.cairographics.org/) ever consider? It would be native on most Linux platforms, may become ISO C++ standard and can be bound to [pango](https://www.pango.org) for platform-independent text handling.
[Skia](https://skia.org/) is a bit heavy and slower on simple tasks.
[GDI+](https://docs.microsoft.com/en-us/windows/desktop/gdiplus/-gdiplus-gdi-start) is terribly slow, designed to paint on windows but not for realtime rendering.

Cyberbeing:
@jesec 
I've pushed my local branch to xy_sub_filter_rc5 but without the VS2019 stuff. It seems basically all of the performance improvement was coming from ad95905, which showed a ~39% improvement on my i5-3570K in UnitTest with your sample script. That said, my results were around 60000ms vs 45000ms compared to your 10225ms vs 7583ms, so I'd be curious if that's all your CPU (which CPU are you running?) or if VS2019, Win10, and the updated platform toolset are actually a factor.

Could you test my VS2013 build and see what performance is like compared to yours?
https://www4.zippyshare.com/v/nfZAKNgG/file.html


> Was [cairo](https://www.cairographics.org/) ever consider?
None of the alternative backends were truly looked into, since the original dev had been planning to write a custom floating-point anti-aliasing scan converter.

jesec:
@Cyberbeing 
As I mentioned in https://github.com/jesec/xy-VSFilter/releases/tag/git-c786186, my test "end" (duration) is 10 (seconds) not default 60. I retested with 60 and my result is 43712 ms. That's not a big difference compared to yours so it is up to you to decide whether to use new platform SDK and toolsets.

However, I am on AMD platform which doesn't implement new instructions like AVX as efficiently as Intel is (Intel brings a 100% AVX perf (-50% cycle per instructions) boost on Skylake and newer).

I personally prefer newer tools as those implemented more state-of-art (C++, concurrency, etc.) features and bring potentially important bug fixes for newer platforms. (especially fixes for potential mis-compilation for new instructions)

Currently I am working on libass as ASS render for XySubfilter. Here is the PoC. https://github.com/jesec/xy-VSFilter/tree/wip-libass-for-ass It is working quite well but it only supports external subs at the moment. Now I am trying to add internal sub support, wire XySubfilter options like time shift and probably add some libass specific options.

I think most of XySubfilter can be solved by using libass for ASS rendering. ASS is probably the only subtitle type which requires any kind of "optimization" as other popular subtitle types like srt (simple dialogue) or pgs (bitmap) doesn't actually require any big resource to handle.

Cyberbeing:
Actually, doing some testing now there seems to be something wrong with ad95905 in my VS2013 build. It's causing extremely high CPU usage and a performance loss on one of my samples https://www97.zippyshare.com/v/wL9QvURE/file.html . CPU usage is ~5-7% with 3.1.0.352 and perf is stable, yet 80-100% with ad95905 which doesn't seem right. If I set everything to num_threads(2) it then seems to instead hover around 33% CPU on my 4 core i5-3570K, but again isn't that really high?

Why is CPU usage increasing linearly with each added thread for real-time playback? I would expect CPU usages to be around the same as single threaded when there isn't a bottleneck. Are the threads looping? Or could this somehow be breaking caching?

@jesec can you reproduce this?

Cyberbeing:
Using Libass to render for XySubFilter is an interesting idea, though I'm curious parts could actually be re-used from XySubFilter when you're done. Are you just doing it this way mainly to re-use XySubFilter's implementation of the subtitle interface?

jesec:
> Actually, doing some testing now there seems to be something wrong with [ad95905](https://github.com/Cyberbeing/xy-VSFilter/commit/ad959054bf5d2fe588911b85a3ea2776c3f8ac62) in my VS2013 build. It's causing extremely high CPU usage and a performance loss on one of my samples https://www97.zippyshare.com/v/wL9QvURE/file.html . CPU usage is ~5-7% with 3.1.0.352 and perf is stable, yet 80-100% with [ad95905](https://github.com/Cyberbeing/xy-VSFilter/commit/ad959054bf5d2fe588911b85a3ea2776c3f8ac62) which doesn't seem right. If I set everything to num_threads(2) it then seems to instead hover around 33% CPU on my 4 core i5-3570K, but again isn't that really high?
> 
> Why is CPU usage increasing linearly with each added thread for real-time playback? I would expect CPU usages to be around the same as single threaded when there isn't a bottleneck. Are the threads looping? Or could this somehow be breaking caching?
> 
> @jesec can you reproduce this?

.\unit_test.exe C:\Users\jc\Downloads\[EveTaku]_Shingeki_no_Kyojin_-_OP_Subtitles01.ass
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from OverallTest
[ RUN      ] OverallTest.xxx
[       OK ] OverallTest.xxx (1006 ms)
[----------] 1 test from OverallTest (1006 ms total)

This script is not a "thing" for the render given 60s can be rendered in 1006ms. 
I did notice a tiny tiny CPU spike when I run unit_test on the scirpt. 
b0f27c47-179a-4f56-a1ab-1a4438a933a3

I think it is quite normal to find it to consume more resources at the a given time frame. Parallelization, unlike optimized algorithms (SSE2 intrinsics, etc.), doesn't decrease the overall works to be done. It just use more of the available resources (cores) to accomplish tasks faster. 

I think there are two things to consider for your situation:
1. How's VS2013's support for OpenMP (parallelization)?
There might be bug in VS2013's implementation?
Or just not as good as VS2019's implementation?
Or even not officially supported in VS2013?

2. My CPU have 16 cores so 3 or 4 threads are not a big deal. But 3 or 4 threads might overload a lower end CPU given decoder and etc. may also need resources. Spawning lots of demanding threads (especially if > available cores) can greatly decrease the system's overall performance. That might be the problem.

To solve those potential problems:
1. Try to upgrade your VS? or use external compilers like ICC or clang. There are plugins to integrate modern compilers with older VS. I am not sure if they support old VS2013 though. I am sure they support 2015 and 2017.

2. There are no real solution to this problem. Implementing a proper concurrency (threading) function with detection of available resources and cores, thread spawning, queuing, etc. is EXTREMELY expensive for this use-case. We can probably use a flag to disable parallelization completely in initialization function if system has too few cores/threads available.

jesec:
Just FYI and side note: Although concurrency/threading is expensive at the micro level (that's why I use OpenMP parallelization), it can be useful at higher level like in RequestFrame. I saw in the SubRenderIntf that RequestFrame() can be async but I don't find any async implementation currently. BTW @madshi , do you know if there is any valid implementation? I tried to implement that (https://github.com/jesec/xy-VSFilter/commit/0febeee7e17e38da608939875a3269c42666c9fc) but unfortunately XySubFilter is just designed for multithreading. Locks are everywhere. Shared data structures, race condition handling etc. are PITA. It is too challenging.

jesec:
> Using Libass to render for XySubFilter is an interesting idea, though I'm curious parts could actually be re-used from XySubFilter when you're done. Are you just doing it this way mainly to re-use XySubFilter's implementation of the subtitle interface and the directshow filter itself?

Most of XySubFilter's input pin handling, subtitle selection, etc. functions can be reused without any problem. I would say it works perfectly. For now I think we only need to hook functions like Open and RequestFrame to open/use libass and we are good to go. I don't experience any problem with those "old" functions.

madshi:
Would it be possible to add libass support in such a way that the user can switch between VSFilter and libass with a simple option? Or would that be too difficult? Such an option would of course be wonderful!

jesec:
> Would it be possible to add libass support in such a way that the user can switch between VSFilter and libass with a simple option? Or would that be too difficult? Such an option would of course be wonderful!

if (m_isass) {
-> if (!m_xy_bool_opt[BOOL_OLD_ASS_RENDERING] && m_isass) {
done. Simple as F. This can also happens at Open function. (don’t make m_isass true at the first place)

May need to restart subtitle renderer. Some simple logics can be easily implemented to do that for us automatically, though. 

Personally I am not really interested in such idea as it could make things complicated in the future like if we want to remove old ASS codes completely. But I can just implement a simple toggle there *for now*. The rest (quick switching) will be left to people who are more interested in this function than I am.

tophf:
There were quite a few things that libass rendered differently.
Not sure switching would be a good idea if those issues are not fixed/addressed.

madshi:
@jesec, of course if the libass solution ends up having no (important) disadvantages, we could remove the VSFilter part of the rendering. But that may require serious testing, and I don't know how quickly end users will be able to detect any potential problems. It could be many weeks or even months until we get enough feedback to know how well libass works compared to VSFilter.

We may end up with a situation where some subtitle files work better with VSFilter. If there's no option to switch back to VSFilter in your libass XySubFilter build, that would result in users having to keep two different XySubFilter builds on their PCs and manually switching the files etc. It's not impossible to do that, but not very user friendly.

@tophf, that's what I'm worried about. Which is why I asked for an option that allows the user to choose VSFilter vs libass. Of course the other solution would be to simply switch to libass and then rely on libass fixing all problems end users find/report. But if libass didn't do that until now, maybe they won't in the future, either, e.g. because they might feel their rendering is "more correct" (although less compatible)?

jesec:
That does make sense. I will probably add a toggle and try not break old codes for now. 

Speaking about ASS subtitles get rendered differently in different implementations. I feel sad about this. We are supposed have a common standard/specification which define which behavior is the right one. I will probably work with ASS standard itself when I have time and if I am interested.

madshi:
Yes, it's unfortunate that VSFilter may not always fully follow the spec/standard, or at least may sometimes have its own private interpretation of the spec. However, this has resulted in many ASS subtitles out there being custom tailored to the specific VSFilter behaviour, which means other renderers have to follow suit, or suffer the consequences. It's a bad position for libass to be in, but there's not much we can do about it now.

Maybe the best way out of this dilemma would be to add yet another ASS header field which describes which rendering engine the ASS file was created for/with? If the header field is missing or says VSFilter, libass would have to do its best to simulate VSFilter behaviour. But if the header field says libass, all weird compatability solutions could be thrown over board.

However, the big question is if we wouldn't just replace one specific interpretation (VSFilter) of the spec with another (libass), with such a header field, for future ASS files.  <sigh>

Cyberbeing:
@jesec I ended up installing VC2019, and ran into the same problem. OpenMP is using ~80% CPU for the sample I posted for the entire time the subtitles are onscreen. In other words it never seems to be completing work, nor making use of the caches to eliminate work entirely. When I tried playing back hi1.ass at 2560x1440, with OpenMP enabled and madVR queue size of 16 and it dropped frames and pretty much pegged CPU at 90-100% throughout, while with OpenMP disabled there were no dropped frames 25-40% CPU usage with the queue never dropped below 10-12. So I really do think it's broken. The UnitTest may be a fluke since it uses VSFilter CSRI with disabled caches, which handles rendering quite differently than XySubFilter.

On an unrelated side note, how are you supposed make VS2019 use vcpkg for the libs you removed, and make those libs static now that the build and project files are also removed? I cloned the vcpkg git, ran bootstrap, ran integration, installed x64 packages via vcpkg, opened the VSFilter.sln with VS2019 IDE but compiling fails saying those includes and libs are missing. Same thing if I try using MSBuild. Same thing if I manually import vcpkg.targets into the project files. While if I manually add the libs and includes which vcpkg built to the project files, the build comes out as dynamic and requires the dll files. What am I doing wrong?

There also seems to be a bug with the VS2019 v142 toolset which seem to be ignoring the cache values in the setting dialog and registry, always using the defaults on my Win7 box. The other setting values seem fine, so I'm not sure what's wrong with that one.

jesec:
@Cyberbeing 
Install x64-windows-static packages.

Cyberbeing:
> Install x64-windows-static packages.
Thanks, that worked. All that time wasted installing boost x64-windows, at least this time I noticed I can install boost-flyweight only (which is all XySubFilter uses) and save quite a bit of time in the process. Other than that, still some compile errors related to "log4cplus/logger.h" in xy_logger.h but adding those VcpkgTriplet Condition lines to basevideofilter.vcxproj & subpic.vcxproj seemed to fix it.

> VSFilter may not always fully follow the spec/standard
Well we've reached the point where VSFilter is spec/standard. Libass has been putting in effort for awhile now to match VSFilter behavior wherever practical, so compatibility isn't as bad as it was years ago. At this point, many of the remaining differences seem to be related to GDI32 vs Harfbuzz in terms of font handling quirks and limitations. Script parsing and rendering behavior, should be pretty close now.

jesec:
> @jesec I ended up installing VC2019, and ran into the same problem. OpenMP is using ~80% CPU for the sample I posted for the entire time the subtitles are onscreen. In other words it never seems to be completing work, nor making use of the caches to eliminate work entirely. When I tried playing back hi1.ass at 2560x1440, with OpenMP enabled and madVR queue size of 16 and it dropped frames and pretty much pegged CPU at 90-100% throughout, while with OpenMP disabled there were no dropped frames 25-40% CPU usage with the queue never dropped below 10-12. So I really do think it's broken. The UnitTest may be a fluke since it uses VSFilter CSRI with disabled caches, which handles rendering quite differently than XySubFilter.

Caches are irrelevant as parallelization happens in lower level (rendering steps of one frame) instead of higher level (frames). As you can see from the codes, with or without parallelization is technically and logically equivalent. 

My test:
script: hi1.ass,
fps = 25,
width=2560
height=1440
start=0
end=60
```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from OverallTest
[ RUN      ] OverallTest.xxx
[       OK ] OverallTest.xxx (17644 ms)
[----------] 1 test from OverallTest (17644 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (17645 ms total)
[  PASSED  ] 1 test.
```
madVR, in XySubFilter resolution set to 2560x1440, no frame drops, queue full (127/128),
a9539878-6c9b-4b5e-b7ca-56a95a2edbc0

> jesec: My CPU have 16 cores so 3 or 4 threads are not a big deal. But 3 or 4 threads might overload a lower end CPU given decoder and etc. may also need resources. Spawning lots of demanding threads (especially if > available cores) can greatly decrease the system's overall performance. That might be the problem.

This may be a valid explanation. Your CPU simply got overloaded due to increased efficiency of resources utilization by XySubFilter. As I said, this can be harmful in some cases. Especially when stuffs like decoder also needs resources. Draining CPU resources may hinder operations of those jobs. 

I suggest you to add breakpoints at relevant parallelization points and check what max iteration count is and see how the tasks got distributed. Especially, check if max iteration count > defined num_threads on your environment. As OpenMP happens in very low level (compiler) for its low overhead characteristics, it could be problematic if OpenMP spawn more threads than max iteration count. Although I carefully evaluated the workload and choose a num_threads optimal to my environment, the situation may be different on your environment.

jesec:
> VSFilter may not always fully follow the spec/standard
> > Well we've reached the point where VSFilter is spec/standard. Libass has been putting in effort for awhile now to match VSFilter behavior wherever practical, so compatibility isn't as bad as it was years ago. At this point, many of the remaining differences seem to be related to GDI32 vs Harfbuzz in terms of font handling quirks and limitations. Script parsing and rendering behavior, should be pretty close now.

This is very true. However, I do firmly believe that things can and should be improved over time. As you can see here https://github.com/libass/libass/issues/328, I have some new ideas and I am trying to propose improvements to ASS standard. But, non-changing (and popular) implementation can make it very extremely difficult to improve a standard and adapt improvements. As libass is progressing while ASS standard development and implementation effort for VSFilter is not active, I believe it is reasonable to switch to libass.

Cyberbeing:
I've always been a big proponent of the idea of introducing a new spec, yet it feels like talks of an ASSv5 spec have been going on for a decade now, with nothing ever getting beyond the planning phase before interest dies off. If you're able to push forward on something new and commit to supporting it on the subtitle renderer side for Windows support, I'd say go for it. The challenge at this point would likely be an authoring app which supports the new spec, since Aegisub isn't really actively developed any more.

> Draining CPU resources may hinder operations of those jobs.

This is my main concern, since I usually equate performance to CPU efficiency. I'm unsure how much sense it makes to use 1600% more CPU power for only a theoretical 40% improvement when that also takes away from decoding resources. What I really need to see is a use-case which definitely drop frames without OpenMP enabled. At that point maybe it would be worth it to enable a couple threads just to get over the real-time threshold. Unfortunately I don't have a 4K display, so it would be nice if something was found where a benefit was reproducible at 2560x1440. Overall though I'd feel better if this was something the user could toggle, if some do see a benefit yet the efficiency impact remains harsh.

> I suggest you to add breakpoints at relevant parallelization points and check what max iteration count is and see how the tasks got distributed. Especially, check if max iteration count > defined num_threads on your environment.

I'm not actually a developer, and have zero experience with multi-threading, so I'm unsure how to actually do that and properly interpret the results.

jesec:
> > Draining CPU resources may hinder operations of those jobs.
> 
> This is my main concern, since I usually equate performance to CPU efficiency. I'm unsure how much sense it makes to use 1600% more CPU power for only a theoretical 40% improvement when that also takes away from decoding resources. What I really need to see is a use-case which definitely drop frames without OpenMP enabled. At that point maybe it would be worth it to enable a couple threads just to get over the real-time threshold. Unfortunately I don't have a 4K display, so it would be nice if something was found where a benefit was reproducible at 2560x1440. Overall though I'd feel better if this was something the user could toggle, if some do see a benefit yet the efficiency impact remains harsh.

I don't think there is any **efficiency** impact there, actually for the subtitle renderer itself the efficiency of resources utilization is much higher. It is just a problem of user's environment. 

Some users like me may have higher end CPU which can easily handle 3 or 4 added threads while others may have lower end ones which can easily get overloaded and thus hinder ops of other jobs. IMO though, most users should have fairly high specs as madVR and relevant video playback stacks are used by more enthusiastic community. Even my laptop with low-voltage CPU (8650U) can quite easily handle this.

As I said, we can implement some kinds of one-time detection at init function and then use the result to decide if parallelization will be enabled. 

If we want to play it safe, I think we can probably change all num_threads to 2.

Cyberbeing:
I played around with it a bit via trial and error, and setting Rasterizer.cpp threads to 2, xy_filter.cpp loop1 threads to 3, and  xy_filter.cpp loop2 threads to 2 seems to be the best balance according to Unit Test with a VC2019 build. Performance loss is only about 5% compared to thread 4/4/3 yet CPU usage reduces ~50% on my i5-3570K. Though this is still around double OpenMP disabled CPU usage on 'easy' cache friendly scripts.

Is there any way to make OpenMP threads sleep for like 1ms between each loop iteration or similar to reduce CPU usage more? Since most of these scripts can be handled decently with a single-thread, I suspect that wouldn't affect real-time playback much. Or I guess the more optimal solution would be if OpenMP could be dynamically enabled/disabled during runtime based on how quickly we need to fill the madVR queue as well of execution time of the loops, but maybe that would be too complex.

If neither of these are practical, then what you said about a one time init at startup would make sense, with possibly a setting to manually override (requiring restart)?

jesec:
> I played around with it a bit via trial and error, and setting Rasterizer.cpp threads to 2, xy_filter.cpp loop1 threads to 3, and xy_filter.cpp loop2 threads to 2 seems to be the best balance according to Unit Test with a VC2019 build. Performance loss is only about 5% compared to thread 4/4/3 yet CPU usage reduces ~50% on my i5-3570K. Though this is still around double OpenMP disabled CPU usage on 'easy' cache friendly scripts.
> 
> Is there any way to make OpenMP threads sleep for like 1ms between each loop iteration or similar to reduce CPU usage more? Since most of these scripts can be handled decently with a single-thread, I suspect that wouldn't affect real-time playback much. Or I guess the more optimal solution would be if OpenMP could be dynamically enabled/disabled during runtime based on how quickly we need to fill the madVR queue as well of execution time of the loops, but maybe that would be too complex.

sleep is not really realistic. OpenMP happens at fairly low level. It is simply, thread 0 do iteration 1, 3, 5, 7, thread 1 do iteration 2, 4, 6, 8. Adding sleep there may damage performance. 
It is possible to instruct OpenMP to use complex scheduling modes but again, like threads/concurrency, the overhead may not worth it. https://www.dartmouth.edu/~rc/classes/intro_openmp/schedule_loops.html

It is also possible to use a variable in num_threads but I am not sure about the overhead. You can probably try:
```
int omp_threads = 3;
if (h < 4) {
omp_threads = 1; // Use single thread if there is not too much work to be done
}
... num_threads(omp_threads)
...
```

Unfortunately madVR has no API to let render know how urgent should the frame be delivered. It is also quite impossible to elevate the load of pending frame before you actually work on it. It might be a simple dialogue or it might be some fancy stuffs. We won't know before we start to work on it. 

> If neither of these are practical, then what you said about a one time init at startup would make sense, with possibly a setting to manually override (requiring restart)?

That's possible.

madshi:
A google search found this:

http://forum.openmp.org/forum/viewtopic.php?f=3&t=1450

So a "passive" OMP_WAIT_POLICY should fix the problem with high CPU usage. Might be worth a try.

Cyberbeing:
Seems that according to this you can set it as Windows environmental variable, 
https://sd.ascon.ru/otrs/public.pl?Action=PublicFAQZoom;ItemID=894

or according to this:
https://ofekshilon.com/2017/11/03/on-omp_wait_policy/

You'd need to set SetEnvironmentVariable(L”OMP_WAIT_POLICY”, L”passive”) somewhere in the DLL. 

Any idea where exactly this should be placed so it actually has an effect in XySubFilter?

The last question would be if the VS2019 preview is supporting it yet. Either way, it probably wouldn't hurt to just set it passive for our use-case even if it doesn't work. I'm a bit tired now, so I'll need to look it after get back up.

jesec:
> The last question would be if the VS2019 preview is supporting it yet. Either way, it probably wouldn't hurt to just set it passive for our use-case even if it doesn't work. I'm a bit tired now, so I'll need to look it after get back up.

VS2019 should support it. 
>> MSDN explicitly mentions (for many versions now) that VC supports only OpenMP 2.0. OMP_WAIT_POLICY is part of the newer OpenMP 3.0 specification, and is the only newer environment variable that MS **implemented**.

@Cyberbeing You can probably test it with env variable first before you try to implement it in XySubFilter.

jesec:
```
set OMP_WAIT_POLICY=passive

unit_test.exe test\hi1.ass
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from OverallTest
[ RUN      ] OverallTest.xxx
[       OK ] OverallTest.xxx (51914 ms)
[----------] 1 test from OverallTest (51915 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (51917 ms total)
[  PASSED  ] 1 test.
```
image

```
set OMP_WAIT_POLICY=active

unit_test.exe test\hi1.ass
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from OverallTest
[ RUN      ] OverallTest.xxx
[       OK ] OverallTest.xxx (43985 ms)
[----------] 1 test from OverallTest (43985 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (43987 ms total)
[  PASSED  ] 1 test.
```
image

jesec:
> You'd need to set SetEnvironmentVariable(L”OMP_WAIT_POLICY”, L”passive”) somewhere in the DLL.
> 
> Any idea where exactly this should be placed so it actually has an effect in XySubFilter?
> 

Unfortunately. It can't be handled by the program itself. 
>> The Visual C++ implementation of the OpenMP standard includes the following environment variables. These environment variables are read at program startup and modifications to their values are ignored at runtime (for example, using _putenv, _wputenv).

Cyberbeing:
For you CPU graphs above is OpenMP multi-threading actually still showing benefit? What do your graphs and Unit Test numbers look like with OpenMP disabled? It seems a bit strange that Core 8 is maxing out. Hopefully that's not just a quirk of AMD Threadripper thread scheduling on Windows. I'll check on my i5-3570K later.

> Unfortunately. It can't be handled by the program itself.

Then I guess the most practical time to set it is when the DLL is registered with RegSvr32?

jesec:
> For you CPU graphs above is OpenMP multi-threading actually still showing benefit? What do your graphs and Unit Test numbers look like with OpenMP disabled? It seems a bit strange that Core 8 is maxing out on Passive rather than load being spread out a bit more. Hopefully that's not just a quirk of AMD Threadripper thread scheduling on Windows. I'll check on my i5-3570K later.

```
[==========] Running 1 test from 1 test case.
[----------] Global test environment set-up.
[----------] 1 test from OverallTest
[ RUN      ] OverallTest.xxx
[       OK ] OverallTest.xxx (59168 ms)
[----------] 1 test from OverallTest (59169 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test case ran. (59169 ms total)
[  PASSED  ] 1 test.
```
image


> > Unfortunately. It can't be handled by the program itself.
> 
> Then I guess the most practical time to set it is when the DLL is registered with RegSvr32?

The check happens each time the DLL is loaded. (in our case, when video player called the DLL) 
To solve this, we probably need to set a persistent environment variable for users in bat.

Cyberbeing:
I checked on my system and seeing the same thing with PASSIVE ~8% perf loss when benchmarking, and one core mostly saturated. During real-time though the core saturation doesn't happen and CPU usage reduces by ~33% from (50-70% to 30-50%) with my 2/3/2 thread build. The Unit Test numbers I'm seeing are pretty much without margin of error of yours.

> To solve this, we probably need to set a persistent environment variable for users in bat.

Same effect either way, since RegSvr32 as well is only ever run once by the user (usually with the bat) on initial install. I do think it would be preferred to have the persistent environment variable written at the same time the DLL is registered with RegSvr32 (since we also write a number of defaults to the registry at this same point), since that makes most sense to ensure defaults are enforced. A bat file would be more practical for when the user wanted to change our chosen default.

jesec:
> I checked on my system and seeing the same thing with PASSIVE ~8% perf loss when benchmarking, and one core mostly saturated. During real-time though the core saturation doesn't happen and CPU usage reduces by ~33% from (50-70% to 30-50%) with my 2/3/2 thread build. The Unit Test numbers I'm seeing are pretty much without margin of error of yours.
> 
> > To solve this, we probably need to set a persistent environment variable for users in bat.
> 
> Same effect either way, since RegSvr32 as well is only ever run once by the user (usually with the bat) on initial install. I do think it would be preferred to have the persistent environment variable written at the same time the DLL is registered with RegSvr32 (since we also write a number of defaults to the registry at this same point), since that makes most sense to ensure defaults are enforced. A bat file would be more practical for when the user wanted to change our chosen default.

```
setx /M OMP_WAIT_POLICY passive
```
does the job.

madshi:
There's also "KMP_BLOCKTIME=infinite/0" and "KMP_LIBRARY=throughtput/turnaround" that seem to be related here. I'm not sure if these are superseded by OMP_WAIT_POLICY, of if they are additional settings to modify the OpenMP behaviour?

Are you using the MS OpenMP implementation or the Intel one? The Intel one seems to have an API named "kmp_set_defaults()" which allows you to "overwrite" the environment variables. No such luck with the MS OpenMP implementation, it seems.

Cyberbeing:
> setx /M OMP_WAIT_POLICY passive

Otherwise I guess you'd need to just write the OMP_WAIT_POLICY to the registry manually if done from within the DLL. You could probably even tie it into the setting dialog and make a toggle for it, but as you mentioned the DLL would be to be restarted for it to take effect.

HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\Session Manager\Environment

> KMP_BLOCKTIME

// OMP_WAIT_POLICY=active should imply blocktime == INT_MAX
// i.e., threads spin-wait forever
// OMP_WAIT_POLICY=passive should imply blocktime == 0

Saw that mentioned in a Clang repository. So I suspect OMP_WAIT_POLICY does superseed that value when present.

jesec:
KMP_BLOCKTIME is a quite new feature (OpenMP 3.0+ or something). MSVC only supports OpenMP 2.0 so we are out of luck. It is a shame that Microsoft does not implement newer OpenMP.

jesec:
set OMP_WAIT_POLICY to passive should solve most problem for us and it is quite safe. I am not sure if we really should implement a toggle for that.

Cyberbeing:
Bit of a roadblock, since I found a sample which loses ~27-33% performance with OMP_WAIT_POLICY PASSIVE compared to OMP disabled or threads 1. OMP_WAIT_POLICY ACTIVE on the other hand showed no performance benefit +/- 1% of OMP disabled or so.

[TnK_ED1.zip](https://github.com/Cyberbeing/xy-VSFilter/files/2854208/TnK_ED1.zip)

So it seems OMP_WAIT_POLICY PASSIVE won't be a good solution if it has a chance of causing a significant performance regression on some scripts. I also noticed during normal playback of that sample it was causing the subtitle queue in madVR to drop by 5-6, which is no good if it's also making real-time playback slower. OMP_WAIT_POLICY ACTIVE & OMP disabled never dropped the queue by more than 1.

I'm beginning to think it would be best to just have OMP threads set to 1 by default with something like a GUI setting for disabled/low/high thread count while retaining OMP_WAIT_POLICY ACTIVE.

jesec:
> Bit of a roadblock, since I found a sample which loses ~27-33% performance with OMP_WAIT_POLICY PASSIVE compared to OMP disabled or threads 1. OMP_WAIT_POLICY ACTIVE on the other hand showed no performance benefit +/- 1% of OMP disabled or so.
> 
> [TnK_ED1.zip](https://github.com/Cyberbeing/xy-VSFilter/files/2854208/TnK_ED1.zip)
> 
> So it seems OMP_WAIT_POLICY PASSIVE won't be a good solution if it has a chance of causing a significant performance regression on some scripts. I also noticed during normal playback of that sample it was causing the subtitle queue in madVR to drop by 5-6, which is no good if it's also making real-time playback slower. OMP_WAIT_POLICY ACTIVE & OMP disabled never dropped the queue by more than 1.
> 
> I'm beginning to think it would be best to just have OMP threads set to 1 by default with something like a GUI setting for disabled/low/high thread count while retaining OMP_WAIT_POLICY ACTIVE.

The example you uploaded does not appear to have any subtitle. 
image

Cyberbeing:
> The example you uploaded does not appear to have any subtitle.
Oops, I uploaded the wrong one. Fixed the attachment now, please re-download it.

The other question what came to mind, is are there any other loops in Rasterizer.cpp or xy_filter.cpp which you thing could benefit from OMP without much effort?

jesec:
> > The example you uploaded does not appear to have any subtitle.
> 
> Oops, I uploaded the wrong one. Fixed the attachment now, please re-download it.
> 
> The other question what came to mind, is are there any other loops in Rasterizer.cpp or xy_filter.cpp which you thing could benefit from OMP without much effort? AdditionDraw?

Unfortunately some of them are C codes (eg. Rasterizer::Rasterize) not SSE intrinsics. Due to limitation of OpenMP 2.0, anything inside omp region can't be optimized (eg. auto convert to SSE/AVX, etc.) by the compiler. It is so expensive to lose such optimization in those C codes. OpenMP 4.0 solved the problem with "#pragma omp simd" but again shame on Microsoft for not support anything beyond OpenMP 2.0.

jesec:
@Cyberbeing 
image

Cyberbeing:
What is that showing, just that your queues are completely stable throughout? What I was seeing is the subtitle queue briefly drop from 127 to 120-124 on some of the heavier transitions when Passive was set watching at 2560x1440, while with Active or OMP disabled it didn't drop below 126.

Running it through Unit Test, both OMP ACTIVE & OMP disabled got ~21500ms, while OMP PASSIVE was ~28500ms.

jesec:
> What is that showing? Just that your queues are completely stable throughout with your CPU using PASSIVE?
> 
> What I was seeing is the subtitle queue briefly drop from 127 to 120-124 on some of the heavier transitions when Passive was set watching at 2560x1440, while with Active or OMP disabled it didn't drop below 126.
> 
> Running it through Unit Test, both OMP ACTIVE & OMP disabled got ~21500ms, while OMP PASSIVE was ~28500ms.

How's your CPU usage?

I think we should be fine if OMP passive doesn't bring some outrageous higher CPU usage than no OMP.
As long as a 60s script can be rendered in 30000ms (also note that we use 4K in unit_test, actual user is likely to use 1080P which cuts the rendering time by a very significant margin, if they do use 4K, you can reasonably expect their specs will be as high as mine, madVR is quite demanding in 4K), there should not be any problem with real-time playback. Yes there might be degradation for this particular case, but sometimes we have to make choices based on the overall gains and losses.

Cyberbeing:
Hmm, I can't actually reproduce the queue drops with Passive in madVR FSE 2560x1440.  Rather than testing at 2560x1440, before I was actually just testing with a maximized window, 2356x1325 active size (0,0,1280x720 -> 102,0,2458,1325), and I can reproduce the queue drops with PASSIVE there. 

I guess the explanation for that could be the XySubFilter scaling load is heavier with a odd scale factor like x1.84 compared to a simple x2.0.

jesec:
> Hmm, I can't actually reproduce the queue drops with Passive in madVR FSE 2560x1440. Rather than testing at 2560x1440, before I was actually just testing with a maximized window, 2356x1325 active size (0,0,1280x720 -> 102,0,2458,1325), and I can reproduce the queue drops with PASSIVE there.
> 
> I guess the explanation for that could be the XySubFilter scaling load is heavier with a odd scale factor like x1.84 compared to a simple x2.0.

That's normal. madVR's downscaling step in after upscaling when window solution is not exactly x(integer). Such processing can be very intensive. I don't think subtitle renderer is the reason for your framedrop here.

Cyberbeing:
Disregard that, tested a bit more and can still reproduce it in FSE 2560x1440, maybe it's somewhat inconstant and one of the non-OMP threads is maxing out.

It's not actually a frame drop, just a brief drop in only madVR's subtitle queue. The other queues are stable so it shouldn't be madVR to blame.

jesec:
> Disregard that, tested a bit more and can still reproduce it in FSE 2560x1440, maybe it's somewhat inconstant and one of the non-OMP threads is maxing out.
> 
> It's not actually a frame drop, just a brief drop in only madVR's subtitle queue. The other queues are stable so it shouldn't be madVR to blame.

I guess that should be fine? If the drops in queue is not significant? By default, madVR uses a 32? CPU queue. I think as long as the queue does not drop 32 (or to be safe 16?) frames, we are good to go.

Cyberbeing:
The main concern would just be users with slower CPUs than I have. My i5-3570K is overclocked to 4.4Ghz, so if someone had an older CPU at stock the impact could be more severe. I just think it may need more testing.

Though for now I guess we could move on with this. Could you add a commit to the pull request which implements:
> If neither of these are practical, then what you said about a one time init at startup would make sense, with possibly a setting to manually override (requiring restart)?

Settings of Disabled, 2/3/2 (low), and 4/4/3 (high) would probably be good. The numbers are for number of threads in RasterizerLoop/XY_FilterLoop1/XY_FilterLoop2.

For the init defaults, probably setting Disabled for 2-Core, Low for 4-Core, and High for 8-Core and above CPUs would be safe when OMP_WAIT_POLICY PASSIVE is set.

jesec:
@Cyberbeing 
It could be complex to use separated thread count variable for all 3 ops so I go the safe option by using 2/2/2 and 3/3/3. I believe setting/option may not be necessarily needed here.

Cyberbeing:
You could probably only change the thread count in Rasterizer.cpp Loop only, as that's the only one which seems to hit full CPU utilization per thread under load. For simplicity, you could probably just use your original defaults in xy_filter.cpp with every preset.

From what I remember, 2/4/3 has identical performance to 2/3/2 as far as I remember (I set it lower only for slightly reduced CPU load), yet 4/3/2 has a performance hit compared to 4/4/3, and under ACTIVE 4/4/4 performs worse than 4/4/3 on my CPU. The ratio between the threads seems semi-important to not create bottlenecks.

I would like an override of some kind though. It doesn't need to be a GUI setting though, Something like a registry key not written by default which overrides the auto thread selection when present would be fine. I think this is important, since I could see a situation where playback of something becomes possible only will 4/4/3 & ACTIVE even though for general use on my i5-3570K I wouldn't want to use that.

jesec:
> From what I read, parallel num_threads overrides omp_set_num_threads when present. So if you used num_threads in xy_filter.cpp, but not in rasterizer.cpp it should work as expected.
> 
> Also, you may as well add a preset for 6 core as well:
> 2 Core = omp_set_num_threads(1)
> 4 Core = omp_set_num_threads(2)
> 6 Core = omp_set_num_threads(3)
> 8+ Core = omp_set_num_threads(4)

@Cyberbeing 
I don't really think we need 4 presets (3 presets should be more than enough) or any kind of user override as they will unavoidably over-complicate the logic and jeopardize the propose of this patch as a drop-in, safe improvement.

Cyberbeing:
Usually I just aim for safe defaults, while having overrides available for quick debugging of issues. 

I still think it would be handy to have a way to turn it off, even if overriding the number of threads were not possible, but I guess we could hold off on that for now until we see if it actually ends up harming real-time video decoding. 4K HEVC decoding in particular can be very heavy if done on the CPU. 

Would it make sense to tweak the thread priority of the OMP threads to Below Normal, or is that even possible?

Also, instead of

```
    int num_procs = omp_get_num_procs();
    if (num_procs > 4)
    {
        omp_set_num_threads(3);
    }
    else if (num_procs == 4)
    {
        omp_set_num_threads(2);
    }
    else
    {
        omp_set_num_threads(1);
    }
```

Could it be simplified to the following?

```
    int num_procs = omp_get_num_procs();
    omp_set_num_threads(num_procs / 2);
```

Or would you be concerned about it scaling to high number of threads on CPUs with lots of cores?

```
    int num_procs = omp_get_num_procs();
    if (num_procs < 8)
    {
        omp_set_num_threads(num_procs / 2);
    }
    else
    {
        omp_set_num_threads(4);
    }
```

In which case for xy_filter.cpp maybe something like this to maintain the thread offset yet still scale the number of threads for optimal performance?

```
int omp_threads = omp_get_num_threads();

//loop1
#pragma omp parallel for num_threads(min(4, omp_threads + 1))


//loop 2
#pragma omp parallel for num_threads(min(3, omp_threads))

```

Thoughts?

jesec:
> @jesec
> Usually I just aim for safe defaults, while having overrides available for quick debugging of issues.
> 
> I still think it would be handy to have a way to turn it off, even if overriding the number of threads were not possible, but I guess we could hold off on that for now until we see if it actually ends up harming real-time video decoding. 4K HEVC decoding in particular can be very heavy if done on the CPU.

Again, if people are using 4K+LAV(software decoding)+madVR, they should have fairly **high** specs. I think things would be fine. 

> Would it make sense to tweak the thread priority of only the OMP threads to Below Normal for extra safety, or is that even possible?

I don't think so.

> Also, instead of
> 
> ```
>     int num_procs = omp_get_num_procs();
>     if (num_procs > 4)
>     {
>         omp_set_num_threads(3);
>     }
>     else if (num_procs == 4)
>     {
>         omp_set_num_threads(2);
>     }
>     else
>     {
>         omp_set_num_threads(1);
>     }
> ```
> Could it be simplified to the following?
> 
> ```
>     int num_procs = omp_get_num_procs();
>     omp_set_num_threads(num_procs / 2);
> ```
> Or would you be concerned about it scaling to high number of threads on CPUs with lots of cores?
> 
> ```
>     int num_procs = omp_get_num_procs();
>     omp_set_num_threads(min(4, num_procs / 2));
> ```

That's a brilliant idea.

> In which case for xy_filter.cpp maybe something like this to maintain the thread offset yet still scale the number of threads for optimal performance?
> 
> ```
> int omp_threads = omp_get_num_threads();
> 
> //loop1
> #pragma omp parallel for num_threads(min(4, omp_threads + 1))
> 
> 
> //loop 2
> #pragma omp parallel for num_threads(min(3, omp_threads))
> ```
> Thoughts?

That would require creation, assignment and sharing of a new variable across different part of the program. I don't think we should make things over-complicated for this.

Cyberbeing:
>That would require creation, assignment and sharing of a new variable across different part of the program. I don't think we should make things over-complicated for this.

There isn't actually a variable being shared, but I guess there is a question as to the scope of `omp_set_num_threads()` and where that modified result is stored. By your reply, you are suggesting OMP stores the result as a local variable, rather than setting it system-wide or program-wide? I ask only since OMP is strange in the respect that I believe it also has a Environmental Variable which does the same thing as `omp_set_num_threads()` meaning it also would have a default when it's not present. I wouldn't be surprised if that call was a system wide override until program is closed, but I don't really know.

_This reminds me that we should probably check that building with logging enabled still works (enabled from vsfilter_config.h) after updating log4cplus from vcpkg. If something has changed in the log4cplus api, I may need your assistance in fixing it's integration into XySubFilter. Adding some debug logging would be the easiest way to find out what OMP is doing exactly. If it does work, how would we go about silencing those log4cplus call Initialize warnings during VC2019 compile which I've never seen before?_
 
If it's not system-wide, I guess `omp_get_num_threads()` would instead pull the default number of threads, or do you think it would just fail? In which case, we'd instead need to do the same thing as the other loop with  `omp_get_num_procs()`?

```
int num_procs = omp_get_num_procs();

//loop1
#pragma omp parallel for num_threads(min(4, (num_procs / 2) + 1))


//loop 2
#pragma omp parallel for num_threads(min(3, num_procs / 2))
```

Other than that could you answer:
> Or would you be concerned about it scaling to high number of threads on CPUs with lots of cores?

You originally set 4 threads for your 16 core CPU, is that because performance was degraded or showed no benefit with 8 threads or some other reason?

jesec:
> > Again, if people are using 4K+LAV(software decoding)+madVR, they should have fairly **high** specs. I think things would be fine.
> 
> I think it's actually the opposite, the people with high-end specs would be unlikely to use LAV software decoding at all for 4K hevc. It's those with systems like mine (i5-3570K, GTX770) which are becoming outdated running a GPU without a hevc decoding asic, which would be stuck using software decoding. The same situation occurred during h264 initial adaption and later on 10bit h264, with xy-VSFilter having very low CPU usage and not interfering with video decoding was what initially made it popular among low spec users yet to upgrade their systems. Back then this was a problem, since many fansubs were released at a single resolution and format, meaning there was no choice if you wanted to watch it from a particular group.
> 

I don't think so. I don't use any kind of hardware decoding methods despite my hardware supports them. I do use software decoding for 4K. Hardware decoding methods are not "no-compromise". It is more likely that hi-end users like me to use software decoding than hardware ones. If they do uses hardware decoding, great, there won't be any CPU consumption at all for decoding part. So it makes idea of subtitle renderer interference even more irrelevant. 

> Alternatively, I guess I could just start providing non-OpenMP builds along with OpenMP builds on the Release page if an override toggle is troublesome to implement in a single build.
> 
> > That would require creation, assignment and sharing of a new variable across different part of the program. I don't think we should make things over-complicated for this.
> 
> There isn't actually a variable being shared, but I guess there is a question as to the scope of `omp_set_num_threads()` and where that modified result is stored. By your reply, you are suggesting OMP stores the result as a local variable, rather than setting it system-wide or program-wide? I ask only since OMP is strange in the respect that I believe it also has a Environmental Variable which does the same thing as `omp_set_num_threads()` meaning it also would have a default when it's not present. I wouldn't be surprised if that call was a system wide override until program is closed, but I don't really know.
> 

There is. Definitely you should NOT call omp_get_num_procs and get your num_procs each time you call functions in Rasterizer or XyFilter. It is better practice to call omp_get_num_procs and get num_procs once and during initialization. The variable is preferably to be a local one instead of one which shares between different part and thus introduce unwanted complexity. 

> _This reminds me that we should probably check that building with logging enabled still works (enabled from vsfilter_config.h) after updating log4cplus from vcpkg. If something has changed in the log4cplus api, I may need your assistance in fixing it's integration into XySubFilter. Adding some debug logging would be the easiest way to find out what OMP is doing exactly. If it does work, how would we go about silencing those log4cplus call Initialize warnings during VC2019 compile which I've never seen before?_

Personally I think those loggings are not useful at all to developers. Breakpoints are much better. Logging are not helpful at all in terms of OMP while breakpoints and parallel watches do help a lot. And I don't have the tools to check if the logging is still working. 

It should work. It is just merely a warning, a reminder. It does not actually check if there is initialization in somewhere. We can safely ignore the warning. 

> If it's not system-wide, I guess `omp_get_num_threads()` would instead pull the default number of threads, or do you think it would just fail? In which case, we'd instead need to do the same thing as the other loop with `omp_get_num_procs()`?
> 
> ```
> int num_procs = omp_get_num_procs();
> 
> //loop1
> #pragma omp parallel for num_threads(min(4, (num_procs / 2) + 1))
> 
> 
> //loop 2
> #pragma omp parallel for num_threads(min(3, num_procs / 2))
> ```

No. I don't think so. And I don't think we should make things more complicated. 

> Other than that could you answer:
> 
> > Or would you be concerned about it scaling to high number of threads on CPUs with lots of cores?
> 
> You originally set 4 threads for your 16 core CPU, is that because performance was degraded or showed no benefit with 8 threads or some other reason?

Creating threads are expensive. It does not worth it to create more than a certain number of threads given the workload. Performance will be compromised.

Cyberbeing:
> Personally I think those loggings are not useful at all to developers.

The log4cplus logging in XySubFilter is primary to provide information about internal values about various aspect of parsing, rendering, loading of subtitles, settings handling, directshow graph connections, subtitle interface connecting, etc. It serves a similar purpose to debug logging in madVR, and can be useful for tracking down bugs. If you don't like physical logs, log4cplus can also output to dbgview.

In any case, building was failing with logging enabled, but thankfully the fix was simple. I only needed to change `log4cplus/logger.h` to `<log4cplus/logger.h>`  in a couple files and add `<log4cplus/loggingmacros.h>` to the includes of `xy_logger.h` and it worked.

After you build with logging, you can enable output by creating a VSFilter.dll.properties or XySubFilter.dll.properties, and place it in the same directory as the dll. The contents of the properties files should contain something like the following for both log & dbgview output, but it can be customized as desired.

```
log4cplus.appender.OUTFILE=log4cplus::RollingFileAppender
log4cplus.appender.OUTFILE.File=${HOMEPATH}\Desktop\XySubFilter.log
log4cplus.appender.OUTFILE.ImmediateFlush=true
log4cplus.appender.OUTFILE.BufferSize=262144
log4cplus.appender.OUTFILE.MaxFileSize=100MB
log4cplus.appender.OUTFILE.MaxBackupIndex=10
log4cplus.appender.OUTFILE.layout=log4cplus::PatternLayout
log4cplus.appender.OUTFILE.layout.ConversionPattern=%D{%H:%M:%S,%q} [%-30l] %-6t %-5p <%x> - %m%n
log4cplus.appender.OUTFILE.filters.1=log4cplus::spi::LogLevelRangeFilter
log4cplus.appender.OUTFILE.filters.1.LogLevelMin=TRACE
log4cplus.appender.OUTFILE.filters.1.LogLevelMax=FATAL
log4cplus.appender.OUTFILE.filters.1.AcceptOnMatch=true
log4cplus.appender.OUTFILE.filters.2=log4cplus::spi::DenyAllFilter

log4cplus.appender.DebugView=log4cplus::Win32DebugAppender
log4cplus.appender.DebugView.layout=log4cplus::PatternLayout
log4cplus.appender.DebugView.layout.ConversionPattern=%D{%H:%M:%S,%q} [%-30l] %-6t %-5p <%x> - %m%n
log4cplus.appender.DebugView.filters.1=log4cplus::spi::LogLevelRangeFilter
log4cplus.appender.DebugView.filters.1.LogLevelMin=TRACE
log4cplus.appender.DebugView.filters.1.LogLevelMax=FATAL
log4cplus.appender.DebugView.filters.1.AcceptOnMatch=true
log4cplus.appender.DebugView.filters.2=log4cplus::spi::DenyAllFilter

log4cplus.rootLogger=All,DebugView,OUTFILE
```

jesec:
> > Personally I think those loggings are not useful at all to developers.
> 
> The log4cplus logging in XySubFilter is primary to provide information about internal values about various aspect of parsing, rendering, loading of subtitles, settings handling, directshow graph connections, subtitle interface connecting, etc. It serves a similar purpose to debug logging in madVR, and can be useful for tracking down bugs. If you don't like physical logs, log4cplus can also output to dbgview.
> 
> In any case, building was failing with logging enabled, but thankfully the fix was simple. I only needed to change `log4cplus/logger.h` to `<log4cplus/logger.h>` in a couple files and add `<log4cplus/loggingmacros.h>` to the includes of `xy_logger.h` and it worked.
> 
> After you build with logging, you can enable output by creating a VSFilter.dll.properties or XySubFilter.dll.properties, and place it in the same directory as the dll. The contents of the properties files should contain something like the following for both log & dbgview output, but it can be customized as desired.
> 
> ```
> log4cplus.appender.OUTFILE=log4cplus::RollingFileAppender
> log4cplus.appender.OUTFILE.File=${HOMEPATH}\Desktop\XySubFilter.log
> log4cplus.appender.OUTFILE.ImmediateFlush=true
> log4cplus.appender.OUTFILE.BufferSize=262144
> log4cplus.appender.OUTFILE.MaxFileSize=100MB
> log4cplus.appender.OUTFILE.MaxBackupIndex=10
> log4cplus.appender.OUTFILE.layout=log4cplus::PatternLayout
> log4cplus.appender.OUTFILE.layout.ConversionPattern=%D{%H:%M:%S,%q} [%-30l] %-6t %-5p <%x> - %m%n
> log4cplus.appender.OUTFILE.filters.1=log4cplus::spi::LogLevelRangeFilter
> log4cplus.appender.OUTFILE.filters.1.LogLevelMin=TRACE
> log4cplus.appender.OUTFILE.filters.1.LogLevelMax=FATAL
> log4cplus.appender.OUTFILE.filters.1.AcceptOnMatch=true
> log4cplus.appender.OUTFILE.filters.2=log4cplus::spi::DenyAllFilter
> 
> log4cplus.appender.DebugView=log4cplus::Win32DebugAppender
> log4cplus.appender.DebugView.layout=log4cplus::PatternLayout
> log4cplus.appender.DebugView.layout.ConversionPattern=%D{%H:%M:%S,%q} [%-30l] %-6t %-5p <%x> - %m%n
> log4cplus.appender.DebugView.filters.1=log4cplus::spi::LogLevelRangeFilter
> log4cplus.appender.DebugView.filters.1.LogLevelMin=TRACE
> log4cplus.appender.DebugView.filters.1.LogLevelMax=FATAL
> log4cplus.appender.DebugView.filters.1.AcceptOnMatch=true
> log4cplus.appender.DebugView.filters.2=log4cplus::spi::DenyAllFilter
> 
> log4cplus.rootLogger=All,DebugView,OUTFILE
> ```

Ack-ed. It does make sense that it may be useful in some cases. Thankfully, you can fix that.

Cyberbeing:
> I don't think so. I don't use any kind of hardware decoding methods despite my hardware supports them. I do use software decoding for 4K.

Either way, this will be something to keep our eye on into the future, so I'd be willing to put adding a disable or override switch on hold until we get reports of issues. 4K hevc fansubs aren't even really a thing yet, so if we're lucky maybe this won't even become an issue.


> No. I don't think so. And I don't think we should make things more complicated.

Okay, so in xy_filter.cpp if keeping the values fixed, looking at the current version of your pull request, you decided to stick with

```
//loop1
#pragma omp parallel for num_threads(4)

//loop 2
#pragma omp parallel for num_threads(3)
```
which seemed to offer the best performance with 4 Rasterizer.cpp threads and while probably a bit overkill at 2 Rasterizer.cpp threads, didn't seem to harm performance other than increasing CPU load slightly?

I guess that make the most sense. If in the future we come across a sample making extremely heavy use of blur, where all threads do max out, we could consider changing it to something lower then.

> Creating threads are expensive. It does not worth it to create more than a certain number of threads given the workload. Performance will be compromised.

Got it.


@jesec 
Almost done now. Sorry for this taking so long.

Last things on the agenda:

* Could you add a commit to set `OMP_WAIT_POLICY PASSIVE` value in the registry at the same time as initial regsvr32? The problem I forsee if we don't this this, is programs bundling XySubFilter and running regsvr32 without realizing `OMP_WAIT_POLICY PASSIVE` should be set. I don't think a BAT file will be reliable enough for this purpose, and the repercussions of a user not realizing `OMP_WAIT_POLICY ACTIVE` is being used by default could be severe.

* Is there a way from our end we could fix installing static lib UnRAR from vcpkg? It seems that now, even though `#USE_UNRAR_STATIC` is set by default, with a 64-windows-static vcpkg installation it builds a dynamic lib which requires unrar.dll despite our settings. If this cannot be fixed without manual intervention, we should probably not use vcpkg for UnRAR and instead look into other options such as adding it as a sub-module instead to keep it up to date.

* XySubFilter currently isn't building fully static with the OpenMP changes. It appears to require VCOMP140.DLL, which I'd like to include as statically linked as well. How would we go about doing that? Or is that not even possible with MSVC? I don't really want to start requiring MSVC runtimes now. If OMP is the problem, are there any other compilers which support OpenMP static linking? If so, how difficult would it be to migrate the XySubFilter project files to support another compile environment? Even if we still support building with MSVC, I'd like some way for the release builds to be fully static.

jesec:
> > I don't think so. I don't use any kind of hardware decoding methods despite my hardware supports them. I do use software decoding for 4K.
> 
> Either way, this will be something to keep our eye on into the future, so I'd be willing to put adding a disable or override switch on hold until we get reports of issues. 4K hevc fansubs aren't even really a thing yet, so if we're lucky maybe this won't even become an issue.
> 
> > No. I don't think so. And I don't think we should make things more complicated.
> 
> Okay, so in xy_filter.cpp if keeping the values fixed, looking at the current version of your pull request, you decided to stick with
> 
> ```
> //loop1
> #pragma omp parallel for num_threads(4)
> 
> //loop 2
> #pragma omp parallel for num_threads(3)
> ```
> which seemed to offer the best performance with 4 Rasterizer.cpp threads and while probably a bit overkill at 2 Rasterizer.cpp threads, didn't seem to harm performance other than increasing CPU load slightly?
> 
> I guess that make the most sense. If in the future we come across a sample making extremely heavy use of blur, where all threads do max out, we could consider changing it to something lower then.

Remember the parallelization happens in micro level. The overall requestFrame (render a frame) process is still single thread. So those parallelizations are independent from each other. The overall maximum threads numbers in any given time frame won't be (num threads of A+B) but (larger of A or B). 

> > Creating threads are expensive. It does not worth it to create more than a certain number of threads given the workload. Performance will be compromised.
> 
> Got it.
> 
> @jesec
> Almost done now. Sorry for this taking so long.
> 
> Last things on the agenda:
> 
> * Could you add a commit to set `OMP_WAIT_POLICY PASSIVE` value in the registry at the same time as initial regsvr32? The problem I forsee if we don't this this, is programs bundling XySubFilter and running regsvr32 without realizing `OMP_WAIT_POLICY PASSIVE` should be set. I don't think a BAT file will be reliable enough for this purpose, and the repercussions of a user not realizing `OMP_WAIT_POLICY ACTIVE` is being used by default could be severe.

I don't really think programs will use regsvr32. They are more likely to load DLL directly without system filter registration. (when they use XySubfilter as internal filter)

> * Is there a way from our end we could fix installing static lib UnRAR from vcpkg? It seems that now, even though `#USE_UNRAR_STATIC` is set by default, with a 64-windows-static vcpkg installation it builds a dynamic lib which requires unrar.dll despite our settings. If this cannot be fixed without manual intervention, we should probably not use vcpkg for UnRAR and instead look into other options such as adding it as a sub-module instead to keep it up to date.

You can compile unrar as static manually and put the libs to vcpkg folders. Or if you can just revert the "remove unrar" part and keep them in thirdparty folder. 

> * XySubFilter currently isn't building fully static with the OpenMP changes. It appears to require VCOMP140.DLL, which I'd like to include as statically linked as well. How would we go about doing that? Or is that not even possible with MSVC? I don't really want to start requiring MSVC runtimes now. If OMP is the problem, are there any other compilers which support OpenMP static linking? If so, how difficult would it be to migrate the XySubFilter project files to support another compile environment? Even if we still support building with MSVC, I'd like some way for the release builds to be fully static.

I don't think that's possible under MSVC (https://visualstudio.uservoice.com/forums/121579-visual-studio-ide/suggestions/3434070-statically-link-openmp-vcomp-dll). Intel C/C++ Compiler can probably do the job. It also supports more recent OpenMP features. However, it is paid and proprietary software and requires additional steps to be integrated. It should not be too difficult to migrate, though.

PR #18, opened 2022-08-28
ASS Format: Introduce LayoutRes{X,Y} script headers

TheOneric:
Proposed in https://github.com/libass/libass/pull/641; see there for details and why. To avoid (further) ASS fragmentation, we (speaking for libass) only want to proceed with the extension if VSFilters also agree. Comments about the headers are also best directed to the libass PR.
The patches here offer an implementation for xy-VSFilter in case you decide to follow the proposal.

Just the first commit already suffices. The second one just changes a function to private and adds some docs to avoid a redundant check+override; if you think keeping both check+overriddes is more robust this commit can just be dropped.

The third and fourth commit have actually nothing to do with `LayoutRes{X,Y}` and can thus be removed at will. They implement an automatic build of both 64bit and 32bit xy-VSFilter and XySubFilter for every future commit via GitHub Actions. Since I don't have a local VS setup, this is what I used to build and test the patches and thought you might find it useful.

---

On another unrelated note, while testing I noticed the CSRI name change mentioned in [3.0.0.306](https://github.com/Cyberbeing/xy-VSFilter/releases/tag/3.0.0.306)’s release notes and implemented in commit 0d64ad7bc03ab26d68150670ef92f3105b24ee20 is not present in `xy_sub_filter_rc5`. I don't see a revert in the git history though. Was there a problem with the name change, or did the commit just slip through by accident?

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,88 @@
+name: Build VSFilter
+
+on:
+  push:
+    branches: [ci, xy_sub_filter_rc*, vsfilter_rc*, master]
+
+jobs:
+  build:
+    runs-on: windows-2019
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - msarch: x64
+            vsfplat: x64
+            namesuf: x86-64
+          - arch: x86
+            vsfplat: Win32
+            namesuf: x86-32
+    defaults:
+      run:
+        shell: 'bash'
+
+    steps:
+      - name: Determine configuration
+        id: config
+        run: |
+          # base name of GHA artifact
+          name="VSFilter+SubFilter"
+          # space delimited list of filename stems to delete from artifact
+          delete=""
+
+          # If it runs for one of the main XySubFilter or xy-VSFilter dev branches,
+          # only publish compatible binaries. If it's some other branch, it's probably
+          # a transient testing build, so keep everything.
+          # ref.: https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1302638201
+          REF="${{ github.ref }}"
+          REF="${REF#refs/heads/}"
+          if [ "$REF" = "master" ] || [ "$REF" = "vsfilter_rc" ] ; then
+              name="VSFilter"
+              delete="XySubFilter"
+          elif echo "$REF" | grep -qE '^xy_sub_filter_rc' ; then
+              name="SubFilter"
+              delete="VSFilter"
+          fi
+
+          echo "name=$name" >> $GITHUB_OUTPUT
+          echo "delete=$delete" >> $GITHUB_OUTPUT
+
+      - name: checkout code
+        uses: actions/checkout@v3
+        with:
+          # We need full history to allow finding the tag in build prep
+          fetch-depth: 0
+
+      - name: install yasm
+        shell: cmd
+        run: |
+          git clone --depth=1 https://github.com/ShiftMediaProject/VSYASM.git
+          .\VSYASM\install_script.bat
+
+      # VS2019 is 16.x;  VS2022 17.x
+      - name: Setup VS2019
+        uses: microsoft/setup-msbuild@v1.1
+        with:
+          msbuild-architecture: matrix.msarch
+      #    vs-version: '[16.01,16.11]'
+
+      - name: Build xy-VSFilter and/or XySubFilter
+        run: |
+          export VS160COMNTOOLS="C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/Common7/Tools/"
+          bash ./build_vsfilter.sh -platform "${{ matrix.vsfplat }}" -compiler VS2019
+
+      - name: Prune incompatible files
+        if: steps.config.outputs.delete != ''
+        run: |
+          for dir in bin/lib*/*/Release ; do
+              for bn in ${{ steps.config.outputs.delete }} ; do
+                  rm -f "$dir"/"$bn".*
+              done
+          done
+
+      - name: Publish output
+        uses: actions/upload-artifact@v3
+        with:
+          name: xy-${{ steps.config.outputs.name }}_nightly_${{ matrix.namesuf }}
+          path: |
+            bin/lib*/*/Release
--- a/src/subtitles/RTS.cpp
+++ b/src/subtitles/RTS.cpp
@@ -1853,14 +1853,10 @@ void CRenderedTextSubtitle::OnChanged()
 
 
 bool CRenderedTextSubtitle::Init( const CRectCoor2& video_rect, const CRectCoor2& subtitle_target_rect,
-    const SIZE& original_video_size )
+    const SIZE& layout_size )
 {
     XY_LOG_INFO(_T(""));
     Deinit();
-
-    // Since this is only called from RenderEx(..), which already overrides
-    // original_video_size with m_layout_size when appropiate, we do not
-    // need to override it again and can just use original_size here
 
     m_video_rect = CRect(video_rect.left*MAX_SUB_PIXEL, 
                          video_rect.top*MAX_SUB_PIXEL, 
@@ -1868,12 +1864,12 @@ bool CRenderedTextSubtitle::Init( const CRectCoor2& video_rect, const CRectCoor2
                          video_rect.bottom*MAX_SUB_PIXEL);
     m_subtitle_target_rect = CRect(subtitle_target_rect.left*MAX_SUB_PIXEL, subtitle_target_rect.top*MAX_SUB_PIXEL, 
         subtitle_target_rect.right*MAX_SUB_PIXEL, subtitle_target_rect.bottom*MAX_SUB_PIXEL);
-    m_size = CSize(original_video_size.cx*MAX_SUB_PIXEL, original_video_size.cy*MAX_SUB_PIXEL);
+    m_size = CSize(layout_size.cx*MAX_SUB_PIXEL, layout_size.cy*MAX_SUB_PIXEL);
 
-    ASSERT(original_video_size.cx!=0 && original_video_size.cy!=0);
+    ASSERT(layout_size.cx!=0 && layout_size.cy!=0);
 
-    m_target_scale_x = video_rect.Width()  * 1.0 / original_video_size.cx;
-    m_target_scale_y = video_rect.Height() * 1.0 / original_video_size.cy;
+    m_target_scale_x = video_rect.Width()  * 1.0 / layout_size.cx;
+    m_target_scale_y = video_rect.Height() * 1.0 / layout_size.cy;
 
     return(true);
 }
@@ -3342,7 +3338,7 @@ STDMETHODIMP CRenderedTextSubtitle::RenderEx(SubPicDesc& spd, REFERENCE_TIME rt,
 
 STDMETHODIMP CRenderedTextSubtitle::RenderEx( IXySubRenderFrame**subRenderFrame, int spd_type,
     const RECT& video_rect, const RECT& subtitle_target_rect,
-    const SIZE& actual_original_video_size,
+    const SIZE& original_video_size,
     REFERENCE_TIME rt, double fps )
 {
     TRACE_RENDERER_REQUEST("Begin RenderEx rt"<<rt);
@@ -3366,10 +3362,10 @@ STDMETHODIMP CRenderedTextSubtitle::RenderEx( IXySubRenderFrame**subRenderFrame,
         cvideo_rect.MoveToXY(0,0);
     }
 
-    SIZE original_video_size = actual_original_video_size;
+    SIZE layout_size = original_video_size;
     if (m_layout_size.cx > 0 && m_layout_size.cy > 0) {
-        original_video_size.cx = m_layout_size.cx;
-        original_video_size.cy = m_layout_size.cy;
+        layout_size.cx = m_layout_size.cx;
+        layout_size.cy = m_layout_size.cy;
     }
 
     XyColorSpace color_space = XY_CS_ARGB;
@@ -3403,9 +3399,9 @@ STDMETHODIMP CRenderedTextSubtitle::RenderEx( IXySubRenderFrame**subRenderFrame,
                                            subtitle_target_rect.top*MAX_SUB_PIXEL,
                                            subtitle_target_rect.right*MAX_SUB_PIXEL,
                                            subtitle_target_rect.bottom*MAX_SUB_PIXEL)
-        || m_size != CSize(original_video_size.cx*MAX_SUB_PIXEL, original_video_size.cy*MAX_SUB_PIXEL) )
+        || m_size != CSize(layout_size.cx*MAX_SUB_PIXEL, layout_size.cy*MAX_SUB_PIXEL) )
     {
-        if (!Init(cvideo_rect, subtitle_target_rect, original_video_size))
+        if (!Init(cvideo_rect, subtitle_target_rect, layout_size))
         {
             XY_LOG_FATAL("Failed to Init.");
             return E_FAIL;
--- a/src/subtitles/RTS.cpp
+++ b/src/subtitles/RTS.cpp
@@ -1857,6 +1857,11 @@ bool CRenderedTextSubtitle::Init( const CRectCoor2& video_rect, const CRectCoor2
 {
     XY_LOG_INFO(_T(""));
     Deinit();
+
+    // Since this is only called from RenderEx(..), which already overrides
+    // original_video_size with m_layout_size when appropiate, we do not
+    // need to override it again and can just use original_size here
+
     m_video_rect = CRect(video_rect.left*MAX_SUB_PIXEL, 
                          video_rect.top*MAX_SUB_PIXEL, 
                          video_rect.right*MAX_SUB_PIXEL, 
@@ -3337,7 +3342,7 @@ STDMETHODIMP CRenderedTextSubtitle::RenderEx(SubPicDesc& spd, REFERENCE_TIME rt,
 
 STDMETHODIMP CRenderedTextSubtitle::RenderEx( IXySubRenderFrame**subRenderFrame, int spd_type,
     const RECT& video_rect, const RECT& subtitle_target_rect,
-    const SIZE& original_video_size,
+    const SIZE& actual_original_video_size,
     REFERENCE_TIME rt, double fps )
 {
     TRACE_RENDERER_REQUEST("Begin RenderEx rt"<<rt);
@@ -3359,6 +3364,12 @@ STDMETHODIMP CRenderedTextSubtitle::RenderEx( IXySubRenderFrame**subRenderFrame,
     {
         XY_LOG_WARN("FIXME: supported with hack");
         cvideo_rect.MoveToXY(0,0);
+    }
+
+    SIZE original_video_size = actual_original_video_size;
+    if (m_layout_size.cx > 0 && m_layout_size.cy > 0) {
+        original_video_size.cx = m_layout_size.cx;
+        original_video_size.cy = m_layout_size.cy;
     }
 
     XyColorSpace color_space = XY_CS_ARGB;
--- a/src/subtitles/RTS.h
+++ b/src/subtitles/RTS.h
@@ -459,6 +459,8 @@ public:
 
 private:
     /*
+     * Expects the original_video_size parameter to already be
+     * overridden by LayoutRes headers when applicable.
      * Will call Deinit()
      */
     bool Init(const CRectCoor2& video_rect, const CRectCoor2& subtitle_target_rect,
--- a/src/subtitles/RTS.h
+++ b/src/subtitles/RTS.h
@@ -457,9 +457,13 @@ public:
     virtual void Copy(CSimpleTextSubtitle& sts);
     virtual void Empty();
 
-public:
+private:
+    /*
+     * Will call Deinit()
+     */
     bool Init(const CRectCoor2& video_rect, const CRectCoor2& subtitle_target_rect,
-        const SIZE& original_video_size); // will call Deinit()
+        const SIZE& original_video_size);
+public:
     void Deinit();
 
     DECLARE_IUNKNOWN
--- a/src/subtitles/RTS.h
+++ b/src/subtitles/RTS.h
@@ -459,12 +459,13 @@ public:
 
 private:
     /*
-     * Expects the original_video_size parameter to already be
-     * overridden by LayoutRes headers when applicable.
+     * The layout_size parameter is either derived from
+     * valid LayoutRes{X,Y} script headers (both must be valid!)
+     * or if those do not exist, equal to the video's storage resolution.
      * Will call Deinit()
      */
     bool Init(const CRectCoor2& video_rect, const CRectCoor2& subtitle_target_rect,
-        const SIZE& original_video_size);
+        const SIZE& layout_size);
 public:
     void Deinit();
 
--- a/src/subtitles/STS.cpp
+++ b/src/subtitles/STS.cpp
@@ -1584,6 +1584,16 @@ if(sver <= 4)   style->scrAlignment = (style->scrAlignment&4) ? ((style->scrAlig
                     : ret.m_dstScreenSize.cy * 4 / 3;
             }
         }
+        else if(entry == L"layoutresx")
+        {
+            try {ret.m_layout_size.cx = GetInt(buff);}
+            catch(...) {ret.m_layout_size = CSize(0, 0); return(false);}
+        }
+        else if(entry == L"layoutresy")
+        {
+            try {ret.m_layout_size.cy = GetInt(buff);}
+            catch(...) {ret.m_layout_size = CSize(0, 0); return(false);}
+        }
         else if(entry == L"wrapstyle")
         {
             try {ret.m_defaultWrapStyle = GetInt(buff);}
--- a/src/subtitles/STS.h
+++ b/src/subtitles/STS.h
@@ -179,6 +179,7 @@ public:
     int            m_defaultWrapStyle;
     int            m_collisions;
     bool           m_fScaledBAS;
+    CSize          m_layout_size;
 
     CSTSStyleMap   m_styles;
 
--- a/src/thirdparty/VirtualDub/Kasumi/Kasumi.vcxproj
+++ b/src/thirdparty/VirtualDub/Kasumi/Kasumi.vcxproj
@@ -39,6 +39,18 @@
     <Import Project="..\..\..\common.props" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x86);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x86);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x64);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x64);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>h;..\h;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>
--- a/src/thirdparty/VirtualDub/system/system.vcxproj
+++ b/src/thirdparty/VirtualDub/system/system.vcxproj
@@ -39,6 +39,18 @@
     <Import Project="..\..\..\common.props" />
   </ImportGroup>
   <PropertyGroup Label="UserMacros" />
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|Win32'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x86);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|Win32'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x86);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Debug|x64'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x64);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)'=='Release|x64'">
+    <ExecutablePath>$(VCInstallDir);$(VC_ExecutablePath_x64);$(CommonExecutablePath)</ExecutablePath>
+  </PropertyGroup>
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>..\h;h;%(AdditionalIncludeDirectories)</AdditionalIncludeDirectories>


Masaiki:
i believe that modifying [here](https://github.com/Cyberbeing/xy-VSFilter/blob/fc01a8da5ea6af9091aaab839bc62dc94a90094e/src/subtitles/RTS.cpp#L3283-L3292) (for xy-VSFilter) and [here](https://github.com/Cyberbeing/xy-VSFilter/blob/fc01a8da5ea6af9091aaab839bc62dc94a90094e/src/subpic/XySubRenderProviderWrapper.cpp#L340-L352) (for XySubfilter) would be better if you want to modify "original_video_size"

TheOneric:
Thanks for the feedback!

Unfortunately it has been a while and I no longer clearly remember why I chose exactly this place. After a quick look I think the following may be why, but please tell me if I got something wrong:
 - it seems to me as if `subpic/XySubRenderProviderWrapper.cpp` doesn't know anything about `STS.h` and `CSimpleTextSubtitle` which has the parsed data values like `m_layout_size`.
  - both the 4- and 7-argument version of `RenderEx` are public functions and the 4-argument version calls the 7-arg one. So it seemed safer to me to make the adjustments at the start of the 7 arg version.
  ofc, if turns out the 7-arg version is never used outside of 4-arg `RenderEx` and also not part of ABI, we could just convert the 7-arg version to `privat` and move the adjustment up to the 4-arg version.
  - needing the `LayoutRes{X,Y}` at just one place seems better than needing to keep two occurences in sync.

Masaiki:
> it seems to me as if subpic/XySubRenderProviderWrapper.cpp doesn't know anything about STS.h and CSimpleTextSubtitle which has the parsed data values like m_layout_size.

the member m_provider of XySubRenderProviderWrapper2 should be able to dynamically convert to RenderedTextSubtitle

Modifying original_video_size is not in line with the semantics of this variable, so the solution I give is to modify the value before it is passed in
Another acceptable solution is
Instead of using original_video_size directly in RenderEx, create a new variable combining m_layout_size and original_video_size

TheOneric:
> Modifying original_video_size is not in line with the semantics of this variable

Is `RenderEx` part of public API? If so, then this ofc necessitates keeping the parameter `const`. I know about VSFilter's `csri_` interface, but I don't really know what is or isn't part of XySubFilter's public API. Checking the exported functions in the DLL doesn't help either as those are just `XySubFilterConfiguration` and some `Dll*` functions. I'm guessing most interesting bits are exported via some function-pointer table object or so.
If you know and could list all external API functions, it would make it easier to figure out what signatures can be tweaked (also for the future). I assume, since `private`ing worked, at least `Init` isn't being exported?

> Instead of using original_video_size directly in RenderEx, create a new variable combining m_layout_size and `original_video_size`

Sounds good in principle and allows to keep the override localised in a single place. Checking closer again, there seem to be only two callss of the 7-arg version of `RenderEx` in this repo outside of `RTS.cpp`. One being the one from `subpic/XySubRenderProviderWrapper.cpp` you linked to before, the other in `subpic/SimpleSubPicProviderImpl.cpp`.
The latter passes `CRect`s which are newly created in the function argument list, so here `const` doesn't matter and there are no `LayoutRes` effects which need to be propagated back. For the former ther could in theory be effects which need to be propagated back.

I can only run xy-VSFilter, but not XySubFilter builds. If I make a GHA testbuild using another patch approach, could you check if XySubFilter still behaves as intended on some testfiles I'll provide? *(To test the current version I also had to enlisted help)*

Masaiki:
> Is RenderEx part of public API? If so, then this ofc necessitates keeping the parameter const. 

No, those public APIs are defined in include/SubRenderIntf.h.

> "originalVideoSize",    SIZE,      info,   read only,  size of the video before scaling and AR adjustments

The values of the three current size-related variables are obtained from the player and are therefore defined as const , which is easy to understand.
Although modifying it(original_video_size) will not cause any problems, we should not modify it for the sake of clear semantics and understanding by subsequent modifiers

Cyberbeing:
> On another unrelated note, while testing I noticed the CSRI name change mentioned in [3.0.0.306](https://github.com/Cyberbeing/xy-VSFilter/releases/tag/3.0.0.306)’s release notes and implemented in commit [0d64ad7](https://github.com/Cyberbeing/xy-VSFilter/commit/0d64ad7bc03ab26d68150670ef92f3105b24ee20) is not present in `xy_sub_filter_rc5`. I don't see a revert in the git history though. Was there a problem with the name change, or did the commit just slip through by accident?

This was intentional, since when XySubFilter development started, the intention was for xy-VSFilter to never be built or used from the XySubFilter dev branches, where breaking changes frequently occurred. Continued xy-VSFilter development would be stable builds only from xy-VSFilter Master. The two branches were being maintained separately.

Simple commits were backported from XySubFilter to xy-VSFilter Master, but xy-VSFilter/CSRI exclusive changes were generally not merged back into XySubFilter. Throughout development of XySubFilter, CSRI broke quite a few times, and some new XySubFilter features have bugs only in xy-VSFilter which never got fixed. The primary reason we kept xy-VSFilter able to built on XySubFilter dev branches, was since it made testing and verification of some things a bit easier.

For this reason, if this change is agreed upon, another pull request will need to be created which is based on xy-VSFilter Master.

astiob:
Very nice to see you again!

Given that there’s no move-fast-break-things development of XySubFilter any more, don’t you think the two branches should be merged into one again? Any bugs ought to be fixed, of course, but keeping two parallel branches only sounds like it doubles the burden.

---

While we have you: over the last year or two people have started to look at pinterf’s GitHub fork as a source of source code and binary builds for XySubFilter and xy-VSFilter, and I suspect (can anyone else confirm?) he builds both from the same commit. Do you have any thoughts on this fork? I got the impression that pinterf only started doing this because you seemed inactive for a while. Personally, I’d love it if we could keep xy-VSFilter and XySubFilter to a single source of truth, no matter which particular source it ends up being. Plus, if the two branches do have meaningful differences and _if_ pinterf is not aware of them, it would be good to explain them.

TheOneric:
Changed the approach to no longer require function signature changes *(but `CRenderedTextSubtitle::Init` is still being privated)*.
My tests with the apparently ill-build VSFilter suggests it still works as intended.
Can somebody please check it behaves as expected for XySubFilter? Here are samples and reference images: [layoutres_demo_02.tar.gz](https://github.com/Cyberbeing/xy-VSFilter/files/9933413/layoutres_demo_02.tar.gz) and binaries build by GHA can be found here: https://github.com/TheOneric/xy-VSFilter/actions/runs/3389585013
*(The reference images were create using libass' pending `LayoutRes` implementation. Thus there's an unrelated slight difference in borderwidth for rotated text, but otherwise it should match)*

---

> This was intentional, [...]

Thanks for explaining. I adjusted the GitHub Action workflow to only remove `VSFilter.*` binaries from artefacts created on any `xy_sub_filter_rc*` branch and (assuming such binaries are ven produced there) `XySubFilter.*` binaries from artifacts created on `master` and `vsfilter_rc` branches. *(btw, which one is the canonical one? currently they point to the same commit)*

Looking at `vsfilter_rc`/`master`, it fortunately already has support f some sort of scaling feature and a 7-argument version of `RenderEx`. The three `SIZE` parameters use different names and going by how they are used are in a different order, but it appears as if `size1` fills the role of `original_video_size`.
Assuming this scaling feature was already fully functional and my assumption about `size1 ` being interchangeable with `original_video_size`, porting the change shouldn't be too hard.
If the two branches can be merged back together this would be even better, but I doubt I'll be able to help in a meaningful way with this task.

---

> I suspect (can anyone else confirm?) he [i.e. pinterf] builds both from the same commit.

I can't say for sure, but at least pinterf's xy-VSFilter binaries do not carry the CSRI renaming I mentioned previously. Which can e.g. be observed in wangqr/Aegisub which bundles pinterf/xy-VSFilter:  https://github.com/wangqr/Aegisub/commit/c0e05562098fca4fccf43790bc7125d11caf5782

TheOneric:
Addendum regarding the VSFilter branch:

What is however going to be a problem for me, is actually building the VSFilter branch. I rely on GitHub Actions for any Windows building. GHA has VS2019 (and newer), but since the last commit to the VSFilter branch is from 2014, this likely won't work and I have no knowledge of VS project files. Either someone with a compatible VisualStudio version will need to build the patches for me to test, or somebody needs to help in getting the branch to build with VisualStudio 2019.

TheOneric:
VSFilter Version: https://github.com/Cyberbeing/xy-VSFilter/pull/19
(as anticipated there are issues with building unrelated to the actual LayoutRes patch)

Cyberbeing:
> Given that there’s no move-fast-break-things development of XySubFilter any more, don’t you think the two branches should be merged into one again? Any bugs ought to be fixed, of course, but keeping two parallel branches only sounds like it doubles the burden.

Yes, this never happened since the original developer @x-xy-y suddenly went missing because of health issues before XySubFilter was feature complete and never returned. My role has always been more of a QC tester and maintainer, not as an actual developer of XySubFilter, so things have just been in limbo with no one to fix bugs.


> While I have you: over the last year or two people have started to look at pinterf’s GitHub fork as a source of source code and binary builds for XySubFilter and xy-VSFilter, and I suspect (can anyone else confirm?) he builds both from the same commit. Do you have any thoughts on this fork?

Yes, I think that is the case. My main thought is still that xy-VSFilter shouldn't really be used from XySubFilter branches, since I cannot in good faith say it's stable and it likely doesn't have pixel identical rendering to either xy-VSFilter Master or XySubFilter. Though as long as any bugs compared to xy-VSFilter Master are fixed when discovered, I don't have any objections.

> Plus, if the two branches do have meaningful differences and if pinterf is not aware of them, it might be good to explain them or perhaps just to merge back his patches into your two branches and continue with them.

Whenever I release a new build here, I do usually go through and merge any bug-fix commits found in either pinterf & mpc-hc branches. I've just been losing motivation to do so these past few years since I've not been watching much fansubbed content.

The main thing which comes to mind feature-wise, is additive blending (fix to avoid sub-pixel gaps) was broken in xy-VSFilter. If that could be fixed, it would be great. I also believe some of the new caches and performance optimizations were never made compatible either. Aside from that, I can't remember. I'd need to dig through some old emails and see if I mentioned them to @x-xy-y.

________________

> master and vsfilter_rc branches. (btw, which one is the canonical one? currently they point to the same commit)

Master is canonical for xy-vsfilter builds.

> somebody needs to help in getting the branch to build with VisualStudio 2019

I just pushed my local branch from a couple years ago which had this change and a few others to master.

________________
It's worth mentioning that idea of adding LayoutRes{X,Y} script headers to the ASS spec is identical a proposal I made years ago during early XySubFilter development, but it ended up rejected after a debate between madVR, Libass, and Aegisub devs at the time.

Unfortunately, at this point I don't remember the exact reasons it was rejected as I was never against it myself, but that would be my only concern. If opinions of those involved have changed over the years and nobody has any objections, I'd still find this useful to implement.

Cyberbeing:
Taking a quick look at the commits, I feel like this may need to be implemented a bit differently.

xy-VSFilter/XySubFilter already has a built-in setting for the user to _manually_ override Layout Resolution under `Renderer Layout Options` in settings. So I think it would make sense for this pull request to use the same method as `Renderer Layout Options - Customize` to override the Layout Res. Namely using `m_xy_size_opt` and disabling (graying out) `Renderer Layout Options` in the UI when the new header exists and is valid. 

Though if we go that route, there may still need to be consideration in regards to automatic BT.601/BT.709 video matrix detention based on `SIZE_ORIGINAL_VIDEO`. So rather than using  `SIZE_USER_SPECIFIED_LAYOUT_SIZE` as with  `Renderer Layout Options - Customize`, we'd likely still need to override  `Renderer Layout Options - Use Original Video Size` `SIZE_ORIGINAL_VIDEO` for it to function correctly.

@TheOneric can you look into this?

astiob:
> in regards to automatic BT.601/BT.709 video matrix detention based on SIZE_ORIGINAL_VIDEO

Now that you bring this up, _actually,_ matrix detection should use information supplied by the demuxer (splitter). I don’t remember if it always exists—and if it doesn’t, then certainly the resolution-based fallback needs to stay—but even when it is, VSFilter currently ignores it.

But when resolution-based guessing is involved, it should continue to use the video’s actual size, not the LayoutRes. It’s only invoked when `YCbCr Matrix: None` (right?) which wants to achieve RGB rendering on the _current_ video.

TheOneric:
Expanding on astiob’s replay regarding colourspace guessing:

With the `csri_` API (is there another public API for VSFilter?), VSFilter only knows the pixelformat, but not the colourspace primaries. Thus guessing remains required here and I too expect using info from the actual video is better and more likely to not mangle colours than guessing based on `LayoutRes{X,Y}`. Btw, do >4K videos also typically use `BT.709 (TV)`? If not perhaps the guess could be expanded now. 
https://github.com/Cyberbeing/xy-VSFilter/blob/8c1fc9c10fa786e4ae51f30924c6ef23b365a3ab/src/filters/transform/vsfilter/csri.h#L55-L101

For XySubFilter, I'm less sure, but comments in `SUbRenderIntf.h` suggest XySubFilter passes unchanged RGB colour values as they were specified in the subs and expects consumers to perform colour mangling as necessary. Meaning even if the code for colour guessing is executed the guessed value would probably never be used. 
https://github.com/Cyberbeing/xy-VSFilter/blob/fc01a8da5ea6af9091aaab839bc62dc94a90094e/include/SubRenderIntf.h#L74-L78
and
https://github.com/Cyberbeing/xy-VSFilter/blob/fc01a8da5ea6af9091aaab839bc62dc94a90094e/include/SubRenderIntf.h#L100-L104

Cyberbeing:
> Now that you bring this up, _actually,_ matrix detection should use information supplied by the demuxer (splitter).

I believe this is already done in xy-VSFilter, where we pull matrix information from the decoder e928f839, while in XySubFilter I believe it is normally pulled from the Consumer yuvMatrix field via SubRenderIntf.h.

The concern would be in regards to matrix detention when unset, set to NONE, otherwise not preset in the source, or overridden to "Guess" in the UI, and if different assumptions need to be made when LayoutRes{X,Y} is present.

The important question for color accuracy, is which video matrix a script with LayoutRes{X,Y} was authored to, compared to the current video matrix, and the adaption needed for correct colors.

I believe we currently implicitly assume that the current unscaled video resolution is the same as the script authoring video resolution for fallback matrix detection. Yet with LayoutRes{X,Y}, this wouldn't make sense to set to anything but the script authoring video resolution. So if LayoutRes{X,Y} doesn't match the current unscaled video resolution, which matrix should be used for correct colors?

Thinking back, this added complexity in niche cases could have been one of the reasons why the LayoutRes{X,Y} script header was never implemented. Though at this point, I'd rather not mess with matrix handling behavior. It feels like it could be a headache for Subtitle Consumers which handle subtitle matrix correction themselves such as madVR if we do.

@astiob How is Libass currently handling matrix niche cases after LayoutRes{X,Y} was implemented?

astiob:
> this is already done in xy-VSFilter, where we pull matrix information from the decoder [e928f83](https://github.com/Cyberbeing/xy-VSFilter/commit/e928f83920b04bb33e0df69c74bbaa8d04a47260)

Oh, that’s interesting! Why the white list though? And it looks like even when the decoder explicitly says the video is BT.601, for big video it’s still overridden by the autodetection of BT.709: https://github.com/Cyberbeing/xy-VSFilter/commit/e928f83920b04bb33e0df69c74bbaa8d04a47260#diff-a90ba887c13062ddfbfe900bc7cd13d6dcecac1ff5c77798c91beda2d23a8425R2266 This doesn’t seem right.

> The concern would be in regards to matrix detention when unset, set to NONE, otherwise not preset in the source, or overridden to "Guess" in the UI, and if different assumptions need to be made when LayoutRes{X,Y} is present.
>
> The important question for color accuracy, is which video matrix a script with LayoutRes{X,Y} was authored to, compared to the current video matrix, and the adaption needed for correct colors.

You _cannot_ know what video matrix the script was “authored to” (in the original public release; even more importantly, in the original editor; etc.). This information simply isn’t available. And `YCbCr Matrix` is defined the way it is defined pretty much precisely to avoid needing video matrices: it says how to convert the subtitle colours to YCbCr in order to put them on YCbCr video, regardless of what that video’s own matrix may be. There is simply no room for LayoutRes to enter the picture:

* If the video is RGB, the subtitles are assumed to be native RGB and no colour conversion happens.

* If `YCbCr Matrix` is set to `None`, the subtitles are likewise—explicitly—native RGB and colour conversion happens only if the subtitles need to be blended onto the (current) video, such that when it is converted back into RGB, the colours are unchanged. That is, using the (current) video’s YCbCr parameters. This is where VSFilter needs to know the video’s matrix or reluctantly guess it if it isn’t known. (And of course, its choice can end up differing from the video renderer’s guess or configuration, but it’s the best VSFilter can do.)

* If the video isn’t RGB and `YCbCr Matrix` is explicitly set to a particular matrix, the subtitles are native YCbCr and need to be decoded using the matrix specified in the header. The video’s YCbCr parameters don’t matter at all.

If you’re thinking of someone taking a subtitle file and reusing it with a differently-encoded video, there’s no fully universal, bulletproof way to guarantee previously-matching colours keep matching, and in the general case a human needs to check them over anyway:

* The video matrices may differ (even at the same resolution), so if the ASS names a `YCbCr Matrix`, it needs to be fixed. (This is part of the original design of `YCbCr Matrix`, which wm4 disliked so much.)
* The ASS may use `None` which _should_ guarantee the colours stay unchanged, but VSFilter may fail to guess the matrix correctly.
* The video itself may have different colours in the first place due to e. g. source differences or plain encoder negligence.

LayoutRes can’t help with any of this.

TheOneric:
One more remark about colour matrices, but if this is pursued further it might be better to move discussion to another issue/pr.

With xy-VSF as DirectShow Filter being able to get colour information (for common video renderers), this leaves mostly just csri without proper `None` support — which is the entrypoint Aegisub uses.
Since csri has extension support baked in, this could be alleviated without breaking API by creating a new `"csri.video_colourspace"` extension and teaching Aegisub about it. Currently xy-VSFilter and guliverkli2-VSF don't implement any extensions. [2007-02-24's asa](https://web.archive.org/web/20070206170044if_/http://asa.diac24.net:80/snapshots/asa-20070204-141955.zip) (the newest snapshot I could retrieve from the internet archive) also doesn't advertise any extensions in `csri_query_ext`, but does recognise `csri.openerr` which can serve as an example. If it was passed as a `csri_openflag` to `csri_open_file` it is used to return the exact error code it got from the system if opening a file failed. The extension's header is located in `csri/include/csri/openerr.h`. 


And a small correction:
> • If the video isn’t RGB and `YCbCr Matrix` is explicitly set to a particular matrix, [...]

The same applies if *no* `YCbCr Matrix` header was set at all by the subs. This case is equivalent to explicitly setting `YCbCr Matrix: TV.601` since guliverkli(2) apparently hardcoded this as the conversion matrix.

TheOneric:
Going back to `LayoutRes*`:

> xy-VSFilter/XySubFilter already has a built-in setting for the user to manually override Layout Resolution under `Renderer Layout Options in settings`. So I think it would make sense for this pull request to use the same method [...]
> can you look into this?

With the pointers from your comment, I believe this approach would amount to introducing a new `SIZE_ASS_LAYOUT_RESOLUTION` entry in `m_xy_size_opt`, which needs to be set to the parsed value at a layer which has access to both the parsed values and `m_xy_size_opt`. And then in eg. (only?) `src/filters/transform/vsfilter/DirectVobSub.cpp` wrapping `XyGetSize( int field, SIZE *value )`’s `switch` over `INT_LAYOUT_SIZE_OPT` in a conditional like this *(a bit different and in `DoGetField` on `xy_sub_filter_rc5`)*:
```C
case SIZE_LAYOUT_WITH:
    *value = m_xy_size_opt[DirectVobSubXyOptions::SIZE_ASS_LAYOUT_RESOLUTION];
    if (value->cx > 0 && value->cy > 0) {
        break;
    } else {
        switch(m_xy_int_opt[DirectVobSubXyOptions::INT_LAYOUT_SIZE_OPT])
        {
        case LAYOUT_SIZE_OPT_FOLLOW_ORIGINAL_VIDEO_SIZE:
            *value = m_xy_size_opt[DirectVobSubXyOptions::SIZE_ORIGINAL_VIDEO];
            break;
        case LAYOUT_SIZE_OPT_USER_SPECIFIED:
            *value = m_xy_size_opt[DirectVobSubXyOptions::SIZE_USER_SPECIFIED_LAYOUT_SIZE];
            break;
        default:
            *value = m_xy_size_opt[DirectVobSubXyOptions::SIZE_ASS_PLAY_RESOLUTION];
            break;
        }
        if (value->cx * value->cy == 0)
        {
            *value = m_xy_size_opt[DirectVobSubXyOptions::SIZE_ORIGINAL_VIDEO];
        }
    }
    break;
```

Notes:

 - `DirectVobSub.cpp` is the only place I could find where the `INT_LAYOUT_SIZE_OPT` is actually acted upon to select a size. The filename suggests though it might only be used for (DirectShow) VSFilter.
 - Going off from the existing handling of `SIZE_ASS_PLAY_RESOLUTION` and not counting initialisation, there are 3(!) places  where `m_xy_size_opt[SIZE_ASS_LAYOUT_RESOLUTION]` needs to be set. `plugin.cpp`, `DirectVobSubFilter.cpp` and `xy_sub_filter.cpp` *(all in `src/filters/transform/vsfilter/`)*. On `master` there are still 2 places. Is one of them used for CSRI and one for DirectShow?
 - Regardless of what all those different paths are used for and from it means triplicating `LayoutRes*` retrieval and hoping this actually covers all entry paths and that theoretical future paths also remember to set that value.

If this looks complete and you want to go with the option-based approach, I'll try to implement it this way. But, personally this seems less appealing to me, it needs code duplication for setting the pseudo-option, requires awareness of all possible entrypoints and how those interact with options and the underlying `STS`/`RTS` parser+render and keeping external user overrides separate from handling of values intrinsic to the format just makes more sense to me.

Regarding greying out `Render Layout Options` in the UI, I’m afraid I have no idea how to do this.
Is greying out even an enhacement for user experience though? I perceived these options more as a workaround for broken files, which should never affect a `LayoutRes` file. And not being able to (pre)set the values just because the current file uses `LayoutRes` seems odd. Perhaps it could be renamed to make it clear it's the “default” or “fallback” for `LayoutRes`-less files?
If it was meant for experimenting with an already running video, greying out might make sense, but I don't immediately see how this would be useful.

---

Unrelated question: which setting select `SIZE_ASS_PLAY_RESOLUTION`? I’ve only seen settings for original video size, PAR-desqueezed video size *(← XySubFilter-only)* and custom layout size in the UI.

TheOneric:
I created option-based implementations to allow a direct comparison, but testing showed there's an issue. CSRI doesn't (have) access to `m_xy_*_options` at all, thus using the option-based approach and setting the pseudo-option at the places mentioned in my previous message works for VSFilter as a DirectShow Filter, but not when VSFilter is used via CSRI.

To fix this `csri_render` can retrieve the parsed value from its `CRenderedTestSubtitle` instance, but the issue lies in passing it back to the renderer. Currently `csri_render` uses `CRenderedTextSubtitle::Render(spd, time, fps, RECToor2& bbox)` and uses the value returned from the last parameter to update an internal value initialised as the frame width and height in `csri_request_fmt` and the following frame will use this value for `spd.vidrect`.¹
Only the 7-argument version pf `RenderEx()` allows passing layout size different from frame size. So fixing this entails either duplicating all code about the `bbox` parameter from `Render()` and 4-arg `RenderEx()` to `csri_render`. Assuming all required info is even accesible from the ooutside this would be roughly 60 lines of code. Or alternatively we could add an additional layout_size parameter to `Render()` and (currently) 4-arg `RenderEx()` and update all other users of those functions.
@Cyberbeing: what would you prefer? *(if you still want to go with the option-based approach that is)*

If someone wants to compare the (incomplete) implementation anyway or perhaps test if at least XySubFilter works as intended (cc @astiob), see here:
**<ins>XySubFilter</ins>**
<dl>
  <dt>Current implementation (confirmed working)</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/commit/3e75a6e8f970e3c609eaac4aae5719e72bf427c2">rc5_layout_res@3e75a6e</a><dd>
  <dt>Option-based alternative (unknown if working)</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/commit/0606e118b2b0fe0c5f0d55f67287a46cfbb6329e">xysf_layoutres_optBased@0606e11</a><dd>
  <dt>Binaries for option-based:</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/actions/runs/3402320261">XySubFilter: LayoutRes OptVersion v1 - GHA Artefacts</a><dd>
</dl>


**<ins>xy-VSFilter</ins>**
<dl>
  <dt>Current implementation (confirmed working)</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/commit/eb170632c4bf6bad8b15677de524aaee4b4a0494">vsfilter_layoutres@eb17063</a><dd>
  <dt>Option-based alternative (works for DirectShow, broken for CSRI)</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/commit/032ec65c340bcdb2115f5daa65f2bd99f98afafb">vsfilter_layoutres_optBased@032ec65</a><dd>
  <dt>Binaries for option-based:</dt>
  <dd><a href="https://github.com/TheOneric/xy-VSFilter/actions/runs/3402255151">xy-VSFilter: LayoutRes OptVersion v1 - GHA Artefacts</a><dd>
</dl>

---

**Footnote:**

[1]: I must admit I’m not even sure why `csri_render_frame` wants to update this parameter or that doing so is correct. From a first look, it seems to me like it should rather reset `video_rect` and `spd.vidrect` to `screen_res` every frame. Since `Aegisub` [re-requests its format before every rendering call](https://github.com/TypesettingTools/Aegisub/blob/8336c7d97cdec30103f301c446d7b970480abbf3/src/subtitles_provider_csri.cpp#L106-L107) this is effectively what happens in Aegisub. 
Though, even if we don't need to update the value, this doesn't make passing a different layout res any simpler, since 7-arg `RendereEx` does return a list of Bitmaps to be blended while `Render` and `RenderEx` blend them onto a provided frame for the caller. *(I'm assuming output rect size is probably the bounding box of the area actually written to)*.
Though maybe there's something I’m missing and perhaps other CSRI users do not reset the format every frame.

@Masaiki iinm you have experience with VSFilter as an AviSynth Plugin. Does the AviSynth codepath use CSRI or the DirectShow filter? If the former, what does it do regarding `csri_request_fmt`?

Masaiki:
As far as I know so far.
All subtitle filters that implement the csri interface only implement the RGB32 format, and all care about only one resolution, which is the resolution of the final display
For more information about xy-VSFilter as a script plugin, see [here](https://github.com/pinterf/xy-VSFilter/blob/xy_sub_filter_rc5/src/filters/transform/vsfilter/plugins.cpp) and [here](https://github.com/HomeOfVapourSynthEvolution/xy-VSFilter/blob/master/src/filters/transform/vsfilter/plugins.cpp)

TheOneric:
> For more information about xy-VSFilter as a script plugin, see [here](https://github.com/pinterf/xy-VSFilter/blob/xy_sub_filter_rc5/src/filters/transform/vsfilter/plugins.cpp) and [here](https://github.com/HomeOfVapourSynthEvolution/xy-VSFilter/blob/master/src/filters/transform/vsfilter/plugins.cpp)

Thanks, so the AviSynth Plugin does neither wrap CSRI nor the DirectShow FIlter, but is it's own entrypoint directly interacting with private functions. Meaning xy-VSFilter binaries expose three different interfaces, XySubFilter binaries (hopefully) only one.

> All subtitle filters that implement the csri interface only implement the RGB32 format,

Well, asa (where csri comes from) did implement all RGB formats and from a superficial lokk behaves more like I'd expect by not (obviously) changing the frame size on its own every frame and already rejecting unsupported formats during `csri_request_fmt` and not only in `csri_render`. But given asa is dead for around 1½ decades already, this isn't too relevant for todays csri users anymore.

More relevant, guliverkli2-VSFilter **did** support all of `BGR_`, `BGR`, `YUV2` and `YV12` as xy-VSF still advertises in `request_format`. xy-VSFilter only supporting `BGR_` now, also has implications for colour accuracy, but this isn't related to `LayoutRes*`. *(Here's [the commit](https://github.com/Cyberbeing/xy-VSFilter/commit/50d3b3b7ab712ee2419f48de14db1b01a8406e43#diff-9f6a5057f9a9f0d9a1079e4a8cf2dc698e56667248d114a4a7e6cd78585372e0) dropping other pixel formats; unfortunately it doesn't elaborate on why exactly)*

TheOneric:
@Cyberbeing, do you have any more questions or concerns?
I voiced my preferred approach, but as I wrote, whatever you decide on, I’ll try to implement.
Regarding the Option Menu I still don't know how to grey something out and couldn't find where the name of the grouped “Layout” options is stored, but we could also change the selectable items, like done [here for xy-VSFilter](https://github.com/TheOneric/xy-VSFilter/commit/fd217e23ad162efe1b7045f8ac537480fc9ce68a).

---

Also:
> It's worth mentioning that idea of adding LayoutRes{X,Y} script headers to the ASS spec is identical to a proposal I made years ago during early XySubFilter development, but it ended up rejected after a debate between madVR, Libass, and Aegisub devs at the time.
> 
> Unfortunately, at this point I don't remember the exact reasons it was rejected as I was never against it myself, but that would be my only concern. If opinions of those involved have changed over the years and nobody has any objections, I'd still find this useful to implement.

I guess this wasn't never answered explicitly yet, but yes all current libass maintainers are supporting a pair of `LayoutRes{X,Y}` headers provided xy-VSFilter agrees. pinterf indicated to follow along with it if there’s an agreement on it and clsid2 was open to looking into it *(but ofc MPC-HC ISR’s scaling is currently incompatible with all other renderers anyway and some things like blur aren't scaled at all in ISR)*.
Aegisub is currently not centrally maintained but split forks of which three (wangqr, TypesettingTools and arch1t3cht) are the most important and would all like to contribute back to upstream when it revives. Their maintainers have varying activity and I doubt they are against `LayoutRes*` as long as someone provides patches (which we are intending to).

As for the [original discussion](https://code.google.com/archive/p/xy-vsfilter/issues/97) there's no clear cut reason why it ended up unimplemented in the end; assuming the comments on the issue are pretty much the complete discussion and not much happened on IRC or somewhere else, I’d speculate it was mainly because the discussion dragged on for a long time and was tiring before reaching an agreement. To give a highly abbreviated recap:
In May 2012 the issue was created suggesting a new header but also generally discussing rendering across resolutions, there was some discussion about how/if XySubFilter already does scaling in a compatible way, how to correctly scale blur and be, etc.
Only in April 2013 the headers specifically were brought to attention again and the names `LayoutRes{X,Y}` suggested. There then was some discussion about naming, whether it should perhaps be a single boolean header to scale with `PlayRes*` or some other variants, how PAR gets involved, some confusion with old libass bugs, and general complaints about ASS quirks which also digressed into some unrelated quirks. Other points brought up included  xy* not being widely used enough yet to introduce a new feature not found in guliverkli2 and whether libass matters. This went on until June, then slowed down.
In January 2014 it was stated the then current libass maintainers were after more discussion on IRC now in agreement to support `LayoutRes{X,Y}` as a header pair. Then silence until April 2014, when a few samples were posted. However, it was pointed out fixing other issues in xy\* had higher priority currently. Around the same time wm4 also opened a PR in libass, which later was closed by accident and then forgotten about: https://github.com/libass/libass/pull/72 .
After April 2014 nothing followed on xy’s tracker.

Tl;dr: Examining the old discussion did not reveal any blockers to me for `LayoutRes{X,Y}` today.

TheOneric:
Apologies for being so annoying; I don’t want this to end up like 2014.

@Cyberbeing, do you have any concerns/questions about the design of the headers themselves or only about the specific way they’re implemented in xy-VSFilter and XySubFilter?
For multiple reasons including me getting more busy for a longer time and upcoming package freezes of some common Linux distributions *(in preparation for their next stable release)*, it would be good if there could be a libass release before December — if possible including `LayoutRes{X,Y}`.
So if the header design itself is uncontroversial but there are still some things needing refinement for xy-VSF/XySF’s implementation, we could already merge the libass implementation earlier *(to ensure `LayoutRes*` will be usable now and not only in some years with the next stable release for those distros)*. Work on this implementation can then continue at a pace comfortable to you.

I do not want to just pass over any input you might have, pressure you to somehow readjust your schedule for this nor just unilaterally publish the header in libass setting the design in stone. So please if you have concerns about the header design itself, want more time to think about it first or generally don't like the headers being released in libass before an xy-VSF/XySF implementation is finished let me know and merging will be deferred.

Masaiki:
@TheOneric I see that you have been invited to become a Collaborator for the repository, which I assume means that this feature has been fully up to you

TheOneric:
Since no objections about the header design were raised, I now merged the libass implementation and will do a release in a couple days.

As for the xy-VSF/XySF implementation, since the last post from Cyberbeing was a request to look into another approach *(which turned out to come with some complications)*,  I’d feel bad to just ignore it after not even a month, so I’ll wait a bit more. But if there’s no further response in some time, I will merge those too.
If someone wants to have `LayoutRes*` in xy-VSF/XySF now already, the current state of both PRs should be fully functional and can be applied as local patches in the meantime. The GHA-built binaries are at the time of writing also still up: [xy-VSFilter](https://github.com/TheOneric/xy-VSFilter/actions/runs/3397971159) and [XySubFilter](https://github.com/TheOneric/xy-VSFilter/actions/runs/3389585013)

> you have been invited to become a Collaborator for the repository, which I assume means that this feature has been fully up to you

Maybe that was the intention, or maybe not; I’m not sure. Being invited as an collaborator also predated the request to consider another approach.

TheOneric:
Some commit message tweaking, otherwise identical to the non-opt-based approach which was confirmed to work as intended before [here](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1302765449).

Since two months passed now, I’ll assume this approach is fine and go ahead with the merge into both branches.

Cyberbeing:
@TheOneric Sorry for disappearing suddenly. I had GPU failure which combined with the holidays and not desiring to pay a huge markup on a new GPU, I was without my dev computer until couple weeks ago.

> Unrelated question: which setting select `SIZE_ASS_PLAY_RESOLUTION`? I’ve only seen settings for original video size, PAR-desqueezed video size _(← XySubFilter-only)_ and custom layout size in the UI.

From what I remember, SIZE_ASS_PLAY_RESOLUTION was an artifact from development which was removed from the UI to avoid user confusion, considering no scripts in circulation were authored to PlayRes only. And those that were would usually have a matching video resolution so it wouldn't matter.

> (Here's [the commit](https://github.com/Cyberbeing/xy-VSFilter/commit/50d3b3b7ab712ee2419f48de14db1b01a8406e43#diff-9f6a5057f9a9f0d9a1079e4a8cf2dc698e56667248d114a4a7e6cd78585372e0) dropping other pixel formats; unfortunately it doesn't elaborate on why exactly)

`CSRI: Nothing change for it. But after checking the code, I believe it is not broken (by me). It should work right with ARGB output. The interface appears to accept outputting other color formats too, but I don't think the original implementation works. So I assume the users of CSRI never using any color format other ARGB.
It uses an old set of interfaces too. (Performance issues TextSub has also happen to it.)`

Looking through my old emails. The above is closest I could find to a stated reason @x-xy-y 


> However, it was pointed out fixing other issues in xy* had higher priority currently. Around the same time wm4 also opened a PR in libass, which later was closed by accident and then forgotten about: [libass/libass#72](https://github.com/libass/libass/pull/72) .

That does sound familiar. I dug through some old emails and it does seem like it was just something that never got implemented before @x-xy-y had to halt development on xy-VSFilter for medical reasons.


> @Cyberbeing, do you have any concerns/questions about the design of the headers themselves or only about the specific way they’re implemented in xy-VSFilter and XySubFilter?

The main concern regarding the existing layout override setting in the UI, was only to ensure it didn't break or conflict with these changes. It ultimately didn't matter which way was implemented as long as that were true.

Now that this has been merged, could someone check to make sure neither `Use Original Video Size` or `AR Adjusted Video Size` user settings in the UI produce unexpected behavior when `LayoutRes{X,Y}` is present? 

Please also check if 'Customize' overrides  `LayoutRes{X,Y}` and still matches behavior of XySubFilter builds prior to the merge.

TheOneric:
Thanks for getting back to this!

> considering no scripts in circulation were authored to PlayRes only.

Such files exist now and the incompatibilities they caused were part of the motivation for the renewed push for `LayoutRes{X,Y}`.
They are afaict mostly the result of automated multi-resolution releases, with the authors and viewers of such releases relying on non-default mpv user-overrides or (old?) MPC-HC ISR behaviour to use `PlayRes` instead of video size. New releases will (hopefully) use `LayoutRes{X,Y}`.

> Now that this has been merged, could someone check to make sure neither Use Original Video Size or AR Adjusted Video Size user settings in the UI produce unexpected behavior when LayoutRes{X,Y} is present?
>
> Please also check if 'Customize' overrides LayoutRes{X,Y} and still matches behavior of XySubFilter builds prior to the merge.

I *think* I tested this before, but I don't remember with 100% certainty anymore. I cannot check now either, since I accidentally deleted my Windows VM and don't have sufficient time to recreate it and then test atm.

I’m fairly certain though, that `LayoutRes{X,Y}` just overrides all other settings, including the options from the UI.Is there any scenario where it makes sense for users to override an explicit `LayoutRes{X,Y}` from the UI? I can't think of one.

Going through the comments here, I [previously](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1312361249) suggested to rename the option to “Layout Defaults” to make it more clear it’s overwritten by `LayoutRes{X,Y}`, but was unable to find where the relevant string to change is located. Do you know where the option group name is stored?
As a last resort if no one knows where it is, we could rename the options in the drop-down menu itself to “Default to ...”, like [here](https://github.com/TheOneric/xy-VSFilter/commit/fd217e23ad162efe1b7045f8ac537480fc9ce68a).

Cyberbeing:
I've not had time to test yet either, but that is mainly since I didn't have any files handy. Now looking at the the Libass issue, it seems you uploaded some at [layoutres_demo_01.tar.gz](https://github.com/libass/libass/files/9398323/layoutres_demo_01.tar.gz) . It may be a good idea to dump the `LayoutRes{X,Y} examples` into the repository /test/ folder later.

Just did a quick test with those files, and `LayoutRes{X,Y}` when present does seem to override all UI options, which is fine. While it may have been nice to keep Customize functional, under the assumption that no `LayoutRes{X,Y} scripts` will have broken layout, it won't matter.

Unfortunately I did notice some unexpected behavior when using `Customize` with the `Render to Original Video Size` checkbox using XySubFilter. While it fortunately doesn't break the LayoutRes{X,Y} specified layout, it appears that it now may be overriding the _subtitle rendering resolution_ rather than the layout.

At least that appears to be what is occurring. I likely won't know for sure until I check a XySubFilter debug logging build (hopefully building those isn't broken at this point). For all I know, this may have just been a bug with that combination of options prior to the merge, which is only more obvious now that LayoutRes{X,Y} is implemented.

The main side-effect is now with  `LayoutRes{X,Y} scripts` if you have `Customize` with the `Render to Original Video Size` enabled, the resolution set there will influence quality and potentially performance of subtitle rendering, which is unintended. 

**I'll just stress to any end-users reading this, that  `Customize` was always a script layout debugging option.
Please switch your default setting to either `Use Original Video Size` _(VSFilter legacy layout compatibility)_ or `Use AR Adjusted Video Size` _(Modern layout compatibility)._**
One of those two options will be accurate for the vast majority of old scripts. `Customize` only exists as a crutch to fix the rare script with broken layout using both options.
**`Customize` was never intended to influence rendering quality.**
_The box to enter a resolution influenced internal layout calculations only._
Now that `LayoutRes{X,Y}` exists, hopefully it will never be needed for any future scripts.

Cyberbeing:
> Going through the comments here, I [previously](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1312361249) suggested to rename the option to “Layout Defaults” to make it more clear it’s overwritten by `LayoutRes{X,Y}`, but was unable to find where the relevant string to change is located. Do you know where the option group name is stored?

https://github.com/Cyberbeing/xy-VSFilter/blob/2a502dc9193aa8ab8d70e33ff16b35b3f575cf3d/src/filters/transform/vsfilter/VSFilter.rc#L377

https://github.com/Cyberbeing/xy-VSFilter/blob/2a502dc9193aa8ab8d70e33ff16b35b3f575cf3d/src/filters/transform/vsfilter/DirectVobSub.cpp#L1145-L1158

https://github.com/Cyberbeing/xy-VSFilter/blob/2a502dc9193aa8ab8d70e33ff16b35b3f575cf3d/src/filters/transform/vsfilter/DirectVobSubPropPage.cpp#L1155-L1183

astiob:
> I did notice some unexpected behavior when using `Customize` with the `Render to Original Video Size` checkbox using XySubFilter. While it fortunately doesn't break the LayoutRes{X,Y} specified layout, it appears that it now may be overriding the _subtitle rendering resolution_ rather than the layout. […] The main side-effect is now with `LayoutRes{X,Y} scripts` if you have `Customize` with the `Render to Original Video Size` enabled, the resolution set there will influence quality and potentially performance of subtitle rendering, which is unintended.

This is precisely the very intent of that checkbox, as reflected in its name, and it has worked this way before LayoutRes.

And to people like me (and [this fellow](https://github.com/libass/libass/issues/656)), it’s great: it avoids [pixel drawing rescaling artefacts](https://code.google.com/archive/p/xy-vsfilter/issues/157) and also prevents subtitles on low-resolution video from [becoming too sharp](https://compare.bakashots.me/compare.php?setId=5364) (although this is less beneficial when using advanced video upscaling, because it appears XySubFilter applies its own bilinear scaling to the subtitles then, which ends up looking worse than the video).

If the checkbox was meant to change the layout resolution, it wouldn’t be separate from the select box. And xy explicitly recommended it as a “workaround” (read: solution) to glitchy pixel drawing scaling (see link above), which implies he did mean it to do exactly this.

Cyberbeing:
> This is precisely the very intent of that checkbox, as reflected in its name, and it has worked this way before LayoutRes.

I think there is a misunderstanding. I know the intent of the checkbox, since I was the one who originally pushed for x-xy-y to implement this feature for reasons similar to what you state.

The original intention of the `Render to Original Video Size` checkbox is to mimic VSFilter.dll style rendering behavior, which always renders and blends subtitles into the video prior to any video scaling. Any deviation from that is a bug, since it is ultimately yet another VSFilter compatibility option.

I do not believe the bug is with `Render to Original Video Size`, but rather `Render to Original Video Size + Customize + LayoutRes{X,Y}`. A quick test showed `Render to Original Video Size` was working correctly with `Use Original Video Size` & `Use AR Adjusted Video Size`. 

The unexpected behavior is with `Customize` which seems to override `Render to Original Video Size` when enabled, breaking the intended functionality. Though it seems quite possible that the underlying cause of this bug existed prior to the merge, it is now more a problem either way on `LayoutRes{X,Y} scripts`.

___
For example, if you set `Render to Original Video Size + Customize 640x360` on the `reused+LayoutRes_from1080pToAnamorphic576p.mkv` test file, the subtitles become a blurry mess. 

If setting `Render to Original Video Size + Customize 7680x4320` the subtitles become extremely sharp and aliased.

In neither case does `Render to Original Video Size + Customize` override `LayoutRes{X,Y}`, which is the only thing the resolution set for Customize should do, if it were doing anything at all. 
___

Considering with `Render to Original Video Size (Disabled) + Customize + LayoutRes{X,Y}`, the resolution set for Customize has no effect, but an untended effect is seen with `Render to Original Video Size (Enabled) + Customize + LayoutRes{X,Y}` I'd consider that a bug.

This suggests that  `Render to Original Video Size + Customize` functionally is broken, and potentially sending an incorrectly sized video frame or frame size information to madVR which does not match `Original Video Size`, breaking intended functionality.

astiob:
Ah, sorry, indeed I misunderstood what you had said. (And thanks for getting the option added in the first place!)

So you’re actually saying that the customized-via-UI size overrides the rendering size when the checkbox is checked? And if I got you right, when LayoutResX/Y is present, the customized size does _not_ override layout, but when LayoutResX/Y is absent, it does?

But maybe this is actually desirable? Currently, you always have the option of using the _actual_ video size by avoiding “Customize”. And actually, I can even argue that it may be desirable to make the checkbox default to LayoutResX/Y as opposed to the physical video size: one of the main claims about LayoutResX/Y is that it allows one to reuse old scripts with new videos by setting LayoutResX/Y to the original video’s resolution. So imagine one takes a script with pixel drawings (or an old script authored for a low-resolution video) and wants to reuse it with a newer video: they add LayoutResX/Y and expect it to keep the old look. This means the pixels in the pixel drawings, as well as the blurriness of the upscaled subtitles, are expected to match the original video (whose size is recorded in LayoutResX/Y), too, not the new video.

Cyberbeing:
> And if I got you right, when LayoutResX/Y is present, the customized size does _not_ override layout, but when LayoutResX/Y is absent, it does?

Correct, with one added detail.

When LayoutResX/Y is present, the Custom size does _not_ override layout, but _does_ override the rendering/output/frame(?) resolution. That last part about overriding resolution is unexpected.

When LayoutResX/Y is _not_ present, the Custom size _does_ override layout, but does _not_ override the rendering/output/frame resolution. This is expected behavior.
</br>
> But maybe this is actually desirable? Currently, you always have the option of using the _actual_ video size by avoiding “Customize”, and this gives you more power as a user. (It would then make sense to rename the checkbox, though.)

I thought about that, but such functionality should be made a separate option if desired, rather than exploiting a bug in Render to Original Video Size which deviates from that option's intention of VSFilter output compatibility.
</br>
> And actually, I can even argue that it may be desirable to make the checkbox default to LayoutResX/Y as opposed to the physical video size

Can you explain what you man by this?
Automatically override Customize set resolution to LayoutResX/Y when present, ignoring the user value?

I'd agree that may be a viable solution. Though somewhat dangerous if someone trolled with authoring a script with 'insane' values for LayoutResX/Y which would likely negatively effect output quality or perf with such an option. Granted if it only occurred when `Render to Original Video Size (Enabled) + Customize + LayoutRes{X,Y}` was set the impact would be minimal.
</br>
> one of the main claims about LayoutResX/Y is that it allows one to reuse old scripts with new videos by setting LayoutResX/Y to the original video’s resolution. So imagine one takes a script with pixel drawings (or an old script authored for a low-resolution video) and wants to reuse it with a newer video: they add LayoutResX/Y and expect it to keep the old look. This means the pixels in the pixel drawings, as well as the blurriness of the upscaled subtitles, are expected to match the original video (whose size is recorded in LayoutResX/Y), too, not the new video.

If you'd like to implement this idea, I approve of you creating a new pull request.

Hopefully at the same time this unexpected behavior with `Render to Original Video Size (Enabled) + Customize + LayoutRes{X,Y}` can be fixed. It sort of feels like we really should just explictly disable the existing Renderer Layout UI options when LayoutRes{X,Y} is present, considering that they already are pretty much disabled and non-functional already aside from this Render to Original Video Size + Customize bug.

The new desired LayoutResX/Y rendering behavior you speak of could be made a new UI option checkbox.

That said, my interpretation of what is going on here with the Customize bug may be incorrect. It seems quite possible this may just be Customize rescaling output rather than rasterizing at a different resolution. It is also possible since the behavior is unexpected, it could result is incorrect scaling of effects such as blur. I've forgotten how it is handled in the code. There is also the question if Subtitle Consumers would handle subtitle frames wildly mismatched from the video resolution gracefully (i.e. hq scaling) as to not degrade quality.

PR #19, opened 2022-11-04
[VSFilter Version]: ASS Format: Introduce LayoutRes{X,Y} script headers

TheOneric:
This is the VSFilter version of #18. Untested as it doesn't build yet. (Thus a draft)

The commit introducing VS2019 support was straightforward enough to cherry-pick and actually even let's the build start. But it fails halfway through with a bunch of errors about undeclared identifiers, wrong types, syntax errors and many more. Illustrative example:
```
D:\a\xy-VSFilter\xy-VSFilter\src\subpic\DX7SubPic.cpp(31,42): error C2065: 'pD3DDev': undeclared identifier [D:\a\xy-VSFilter\xy-VSFilter\src\subpic\subpic.vcxproj]
```

Fixing them is beyond my very limited MSVC knowledge (it requires knowing why and how it worked with older versions). If someone wants to take a stab at it and logged-in on GitHub, then see [here](https://github.com/TheOneric/xy-VSFilter/actions/runs/3390470903/jobs/5634653655#step:6:157) for the errors. Otherwise here's an archive with the full logs for the GHA run: [logs_55.zip](https://github.com/Cyberbeing/xy-VSFilter/files/9934176/logs_55.zip).
Alternatively, if someone with VS2013 could build the patched source and upload a binary for testing, this would also work for now to at least figure out if it's even the right approach.

Note: If wanting to verify the CRLF-changes in the GitHub UI, there's an “Hide whitespace” option available in the cog menu located over the commit message.


Masaiki:
just remove DX7SubPic.cpp and DX7SubPic.h from the project
the rest problems are easy to solve

TheOneric:
After rebasing on the updated `master` branch with VS2019 support (thanks!), it [now builds](https://github.com/TheOneric/xy-VSFilter/actions/runs/3397971159) and nightly artifacts work as well. Still haven't tested though; waiting on the decision whether the approach should be switched out for an option-based one.

TheOneric:
Following the merge of the XySubFilter version, this too will be merged and the preceding force-push just updated the commit messages.

madshi:
Is there anything that I need to do on my (madVR) side, to support the new changes you guys are doing? I'm very low on time atm, but I can put it on my to do list.

astiob:
I _think_ not; thanks for asking though!

This adds new headers within the subtitle stream that give the subtitle renderer information that it otherwise has to guess or obtain from madVR or DirectShow, and it affects only the subtitle renderer’s internal computations. If the new headers are absent, nothing changes. If they’re present, madVR still acts the same as before but XySubFilter sees the headers and uses them in lieu of the `originalVideoSize` it gets in `ISubRenderOptions`.

madshi:
Sounds good. If you do anything in the future that needs changes in madVR, could you please notify me at madshi (at) gmail (dot) com? Thanks!

PR #20, opened 2022-11-05
fix: filter operation when changing play rate.

Masaiki:
when I speed up the video, the internal subtitles of mkv are doubly accelerated.

fixes port from https://github.com/v0lt/VSFilter/commit/13a2b5367bde8f6935b08618c3f56b1a249c13c9 and https://github.com/v0lt/VSFilter/commit/94f82394c265bce7f57fba703d67b176506cf93c


Cyberbeing:
Are you aware off hand if std::find is supported by VS2010? Hopefully my old build env is still functional to check.

Cyberbeing:
I asked ChatGPT and it stated:
`Yes, the Visual Studio 2010 compiler supports std::find. std::find is part of the C++ Standard Library, which was included in C++98, and has been available in Visual Studio since its 2010 version.`
So I'll assume it is fine and merge.

PR #21, opened 2022-11-05
fixes about PGS subtitles

Masaiki:


Masaiki:
calling for review @Cyberbeing

clsid2:
These fixes are coming from MPC-HC and are good.

If possible also have a look at porting WebVTT support from MPC-HC to both your projects.

Cyberbeing:
Since I'm aware that external PGS support in xy-VSFilter has been buggy, I'll assume these changes are good and go ahead and merge.

PR #23, opened 2023-01-05
VSFilter: misc colour changes

TheOneric:
Not actually tested yet, but putting this up so it’s less likely to be forgotten about.

@astiob: is the change in the 3rd commit what you had in mind [here](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1304113232)?

The idea behind officially supporting BGRA is that it could be used for colour-accurate rendering with CSRI by allowing the CSRI user to perform colour mangling like they’d do with libass. Without an alpha channel this isn't possible without degrading the video’s colours.
I think aegisub already uses/expects the alpha channel to have ASS semantics where 0xFF is fully transparent and 0x00 fully opaque, but this should be rechecked and it might also be good to check what alpha semantics asa’s CSRI interface used/expected.


src/filters/transform/vsfilter/DirectVobSubFilter.cpp
```diff
@@ -2280,10 +2280,19 @@ void CDirectVobSubFilter::SetYuvMatrix()
...
-                          extformat.VideoTransferMatrix == DXVA2_VideoTransferMatrix_SMPTE240M ||
-                          m_xy_size_opt[SIZE_ORIGINAL_VIDEO].cx > m_bt601Width ||
-                          m_xy_size_opt[SIZE_ORIGINAL_VIDEO].cy > m_bt601Height) ? ColorConvTable::BT709 : ColorConvTable::BT601;
+            switch (extformat.VideoTransferMatrix)
+            {
+            case DXVA2_VideoTransferMatrix_BT709:
+                yuv_matrix = ColorConvTable::BT709;
+                break;
+            case DXVA2_VideoTransferMatrix_SMPTE240M:
+                yuv_matrix = ColorConvTable::BT601;
```
astiob, line 2289: SMPTE 240M and BT.601 have different matrices.
This differs from the original code, and neither seems correct.

See: https://learn.microsoft.com/en-us/windows/win32/api/dxva2api/ne-dxva2api-dxva2_videotransfermatrix

astiob:
> @astiob: is the change in the 3rd commit what you had in mind [here](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1304113232)?

Probably. But the commit doesn’t actually change anything when the upstream says the matrix is BT.601.

TheOneric:
New version:
 - BT.601 is now actually preserved
 - SMTP240M gets conflated with BT709 again for now, since `ColorConvTable` [doesn’t appear to support SMTP240M](https://github.com/Cyberbeing/xy-VSFilter/blob/2a502dc9193aa8ab8d70e33ff16b35b3f575cf3d/src/subpic/color_conv_table.h#L16-L19) *(this should ideally also be fixed)* and in fact on [current master](https://github.com/Cyberbeing/xy-VSFilter/blob/825c27cce4b4dadfb30666ffd547d1bcdde19002/src/subpic/color_conv_table.h#L16-L19) `BT2020` is also missing.
  - now all source files with at least one preexisting CRLF are converted to full CRLF instead of only the files touched in this PR.

src/filters/transform/vsfilter/DirectVobSubFilter.cpp
TheOneric: Thanks! I probably mixed up SMPTE240M and Bt601 while writing. Should now be fixed, though actual SMPTE240 isn't supported by ColorConvTable (yet) so it gets conflated with BT709 again (for now) like in the original code.

TheOneric:
Copying Cyberbeing’s [comment from another PR](https://github.com/Cyberbeing/xy-VSFilter/pull/18#issuecomment-1416136665) for future reference:
> > (Here's [the commit](https://github.com/Cyberbeing/xy-VSFilter/commit/50d3b3b7ab712ee2419f48de14db1b01a8406e43#diff-9f6a5057f9a9f0d9a1079e4a8cf2dc698e56667248d114a4a7e6cd78585372e0) dropping other pixel formats; unfortunately it doesn't elaborate on why exactly)
>
> `CSRI: Nothing change for it. But after checking the code, I believe it is not broken (by me). It should work right with ARGB output. The interface appears to accept outputting other color formats too, but I don't think the original implementation works. So I assume the users of CSRI never using any color format other ARGB. It uses an old set of interfaces too. (Performance issues TextSub has also happen to it.)`
>
> Looking through my old emails. The above is closest I could find to a stated reason @x-xy-y

Cyberbeing:
One thing to mention when making changing to the color handling code, is to be careful to ensure BT.601 continues to always be used on `Auto` when `YCbCr Matrix:` is _not_ present in a script. That is a compatibility behavior for legacy scripts, since VSFilter 2.39 supported BT.601 only.

TheOneric:
btw, how about merging the first two or only the first commit(s) early?

The first one only changes whitespaces to make line endings consistent in each file and should thus be totally safe. Since it touches many files, leaving it unmerged makes future merge conflicts likely.

The second commit seems like a strict improvement to me, since before `csri_request_fmt` pretended like non-BGR formats were supported, but then during rendering crashed (with `assert`) or invoked UB and (tried to) return without actually rendering. Thus it seems to me like no one could use those formats anyway. With the commit applied `csri_request_fmt` already tells potential users that the format is not supported, so format probing to select the best choice is possible again.

PR #25, opened 2023-02-07
[VSFilter] gui: clarify LayoutRes{X,Y} overriding GUI setting

TheOneric:
The change seems rather innocous, but I guess it would be good if someone can actually test it before merge (I cannot atm). Binaries with this patched are available here: https://github.com/TheOneric/xy-VSFilter/actions/runs/4118925204

Unlike in XySubFilter *(see #26)*, there seems to be no occurence or check against the old `"Use Original Video Size"` name anywhere else in the codebase.


Cyberbeing:
I would suggest at this time simply changing `Renderer Layout Options` to `Renderer Layout Default` only.

For both xy-VSFIlter & XySubFilter that would have the least impact to the UI, and saved user settings in the registry.
registry
If the setting names were changed, existing  `Renderer Layout Options (layout_size_opt)` settings would be ignored until the UI is opened and settings re-saved. If we specified a new settings version number, the new default settings could be re-written, but that is unfavorable since in our current implementation of settings versioning it may overwrite all user settings to defaults, not just what we changed.

For xy-VSFilter, renaming `Use Original Video Size` to `Original Video Size (vsfilter2.39)` is redundant. Since from the perspective of xy-VSFilter that is not a VSFilter 2.39 compatibility option, but rather forced behavior as a result of being a DirectShow Transform filter and implicitly known. We likely kept it named that way for consistency with XySubFilter, but for xy-VSFilter it could have very well just been named `Default`. Specifying further only really makes sense for XySubFilter which has alternative possibilities.

Even with XySubFilter #26, specifying VSFilter 2.39 in particular doesn't make much sense, rather than just `Original Video Size (VSFilter)`. Though as mentioned, it may be best left alone for the time being.

If we do end up renaming the combo box options, keep in mind that working with windows .rc resource files is painful. If the string length is increased, you'd also need to increase the size of the combo box and possibly re-position the H: and W: boxes manually:
cutoff

TheOneric:
Thanks for the feedback and testing!

With the name change for the drop-down options not being cosmetic only, but either troubling users or needing additional logic to convert old names, I agree it’s more hassle than its worth.
New version only renames the group label to `Renderer Layout Default` as suggested. This should still make `LayoutRes*` having higher prio sufficiently clear.

TheOneric:
> [Cyberbeing:] I would suggest at this time simply changing Renderer Layout Options to Renderer Layout Default only.

Since the new version now exactly matches what was proposed 2 weeks before, I’ll assume it is fine and merge it + another small commit to include `License.txt` in nightly build archives.

PR #26, opened 2023-02-07
[XySubFilter] gui: clarify LayoutRes{X,Y} overriding GUI setting

TheOneric:
Different from VSFilter’s #25, XySubFilter’s code has several occurences and checks against the old `"Use Original Video Size"` and `"Use AR Adjusted Video Size"` names in `DirectVobSub.cpp`. Afaict all occurrences were updated.

@Cyberbeing are those strings somehow visible to the outside? If so, we should probably leave them uncahnged and only update `Renderer Layout Options` → `Renderer Layout Res Default`.

As with #25 I cannot test this myself atm, if someone wants to verify this doesn't break anything, here are binaries: https://github.com/TheOneric/xy-VSFilter/actions/runs/4118933110


TheOneric:
[Turns out](https://github.com/Cyberbeing/xy-VSFilter/pull/25#issuecomment-1422016587) the option names do indeed leak to the outside in form of registry strings. So renaming them comes with more hassle than it’s worth.
New version only changes the label of the option group to `Renderer Layout Default`.

TheOneric:
Same as https://github.com/Cyberbeing/xy-VSFilter/pull/25#issuecomment-1437003369

